PR #73, opened 2016-10-06
Integrate Tangram and OnTheRoad frameworks, and build samples for each.

### LINKED ISSUE #58: Build sample for routing

msmollin:
Fixes #58. This culimates the work of integrating Tangram-es and On The Road for iOS. Samples have been built out showcasing how to use each framework. Future iterations will continue to hone this approach.


Podfile.lock
```diff
@@ -1,16 +1,42 @@
 PODS:
-  - Pelias (1.0.0-beta1):
-    - Pelias/Core (= 1.0.0-beta1)
-  - Pelias/Core (1.0.0-beta1)
+  - OnTheRoad (0.1)
```
ecgreb: Is the .lock file usually committed to source control?

msmollin: Yeah. It's a recommended thing these days. http://guides.cocoapods.org/using/using-cocoapods.html#should-i-check-the-pods-directory-into-source-control

ios-sdk/TangramVC.swift
```diff
@@ -0,0 +1,35 @@
...
+//  TangramVC.swift
+//  ios-sdk
+//
+//  Created by Matt Smollinger on 7/12/16.
+//  Copyright © 2016 Mapzen. All rights reserved.
+//
+
+import UIKit
+import tangram_framework
+class TangramVC:  TGMapViewController{
```
karimnaaji: It's a little odd to mix up `TangramVC` with other names like `RoutingViewController`, they should be named following the same pattern. I personally like them explicit (leaving `ViewController`) for better understanding.

msmollin: Agreed. That's just sloppiness on my part. Will refactor in next PR.

PR #78, opened 2016-10-17
Update Readme for installation instructions

### LINKED ISSUE #21: Add README

msmollin:
Fixes #21


ecgreb:
@tallytalwar can you test drive these setup instructions and report back if you run into any issues.

Looks like the main challenges might be around getting Cocoapods installed correctly for your specific version of xcode/swift.

karimnaaji:
Instructions looks good!

One issue I had after running `pod install`:

screen shot 2016-10-24 at 2 29 58 pm

When looking at the provisioning Profile
screen shot 2016-10-24 at 2 31 02 pm
:

What are the steps to get it to build in that case?

msmollin:
Ah yeah should probably put a note in there if you're building for device (which is required until we get simulator support working in tangram) that you have to change the bundle identifier to something you can provision for, and then go through all the steps of making a provisioning profile for it.

karimnaaji:
I see, those steps/instructions would be a good addition to build for device.

We already added simulator support in Tangram from https://github.com/tangrams/tangram-es/pull/1003, we only need to move forward and update the binary framework.

tallytalwar:
BTW, @karimnaaji got simulator added to ios-framework.

msmollin:
@karimnaaji / @tallytalwar added a note about the code signing issue in the readme. 

We can update the readme again once we're ready to merge in simulator support from tangram.

karimnaaji:
Looks good.

PR #90, opened 2016-11-11
Updates for Single API key and minor cleanup

### LINKED ISSUE #67: Single API key

msmollin:
Also includes updates to the bubble wrap 1.0 scene which hopefully we can remove soon for a more elegant way of loading house styles.

Fixes #67


msmollin:
The current problem I'm running into is I can't seem to figure out how to format the API key string correctly. This results in tangram not loading a map in the SDK demo app. I've tried both single line YAML syntax and the current dot notation style syntax. Neither of these options produces a functioning map in the demo application.

ios-sdk/TangramVC.swift
```diff
@@ -12,6 +12,7 @@ class TangramVC:  TGMapViewController{
 
     override func viewDidLoad() {
         super.viewDidLoad()
+        self.queueSceneUpdate("sources.mapzen.url_params.api_key", withValue: "vector-tiles-7bfyfz8")
```
karimnaaji: Two things:
- I don't see any scene loaded anywhere?
- `self.applySceneUpdate()` must be called somewhere for this update to be effective.

For future reference, you can see here what is being triggered once a scene gets loaded: https://github.com/tangrams/tangram-es/blob/master/core/src/tangram.cpp#L182, most important thing to notice is that scene updates get cleared up at any `loadScene*` call.

blair1618: Looking at https://github.com/mapzen/ios/blob/master/tangram/scenes/scene.yaml#L384-L393 I think the issue is that we need to update to the latest version of bubble-wrap, which has sources that look like this https://github.com/tangrams/bubble-wrap/blob/gh-pages/bubble-wrap.yaml#L537-L548 - after that this line should work perfectly!

blair1618: Ach, disregard my comment above! I hadn't noticed that this branch also updates the scene file. 

Another problem could be that scene updates don't create all necessary nodes along the path to a target node. In other words, it acts like `mkdir` and not `mkdir -p`. Since the version of bubble-wrap here doesn't already have a `url_params` node, you could try to make this update something like:
```
queueSceneUpdate("sources.mapzen.url_params", "{ api_key: xxxxxx }")
```

msmollin: Previously I was relying on the scene load that Tangram is does automatically if you don't specify a scene file; basically I have the `scene.yaml` placed in the correct spot by Xcode during the project copy. However that's bad form I'd imagine so I'll update to call explicitly from code.

As for the `applyUpdate()` method not being invoked - I don't see that exposed in the public header here - https://github.com/tangrams/tangram-es/blob/master/ios/src/TGMapViewController.h

Am I missing something?

karimnaaji: I see, previously this was hardcoded, it has been removed since a while back from `TGMapController` within tangram-es. We don't want the user to load that `scene.yaml` by default (https://github.com/tangrams/tangram-es/commit/97062b84150db5736b347492f8e9426824c6b8bb#diff-6e7af98500d6b23bde5e916d3e8dc4f3L353) but instead use `loadScene` explicitely, so we don't have to carry on that `scene.yaml` resource while distributing the framework.

As for the `applyUpdates()`, it's definitely missing from the interface. Pushing a short fix shortly!

karimnaaji: Added the missing piece https://github.com/tangrams/tangram-es/pull/1079

karimnaaji:
From looking at the code this looks good to me, but I have several concerns:

- Can we monitor and make sure that the requests have the correct API key from the network requests?
- We might be missing some appropriate external sdk helpers to setup this API key since it's now hardcoded, it seems that it would be great to leave this configurable from outside the SDK (In the future for the scene too, but this seems to be out of scope of this PR).
- It doesn't seem that we are using a single API key, from my understanding we want to be moving to using one API key for all the services? Pelias/Vector Tile etc. In this PR we would still be using one API for each of the services.

msmollin:
@karimnaaji updated to have correct single API keys and retested with new updates from Tangram-es and all code is working correctly

In a future PR I will be adding those helpers for setting API keys easier, per our discussion.

screen shot 2016-11-15 at 4 48 41 pm

karimnaaji:
Nice! Looks good to me!

blair1618:
I'm unclear why we are putting an API key in the SDK at all. I would understand putting an API key in a demo app, but shouldn't a user get their own API key when they incorporate the SDK in their app?

karimnaaji:
Yes, I guess the main question here is whether or not we enforce usage of a new API key when you start using the SDK (one advantage of that is easily reaching out to devs and more usage stats).

msmollin:
One thing to note - this is actually the demo app code. The wrapping code is coming soon which will enforce API key requirements for the individual services. Still working out the best architecture for that.

blair1618:
Oh ok, so it sounds like the problem I'm really pointing out is that there is currently no separation between the SDK and the demo app - and we are aware of this problem already.

Issue #31, opened 2016-06-14
Implement find me button to recenter the map


msmollin:
- [x] Display current location gem on map at fixed position at launch.
- [x] Use real device location to position current location gem on app launch.
- [x] Make current location gem follow device location with real location updates.
- [x] Add find me button. Reset position, zoom, and tilt on click.
- [x] Toggle tracking on/off by adding activated/deactivated state to find me button.
- [x] Add public method to enable/disable the entire current location layer.

PR #91, opened 2016-11-21
Basic Location Gem addition

msmollin:
Very rudimentary usage of location gem - just trying to get displayed. Future PR will clean this up to make it more functional / testable. 

First part of work for #31

### DIFF
--- a/ios-sdk/src/MapViewController.swift
+++ b/ios-sdk/src/MapViewController.swift
@@ -13,52 +13,39 @@ import TangramMap
 public class MapViewController: TGMapViewController {
 
     var currentLocationGem: TGMapMarkerId?
+    //TODO: This will eventually get refactored out to be whatever the current location is from CoreLocation
+    let tempPoint = TGGeoPoint(longitude: -122.44880676269531, latitude: 37.76155490343394)
 
-    public var showCurrentLocation: Bool = false {
-        didSet {
-            if showCurrentLocation {
-                let point = TGGeoPoint(longitude: -122.44880676269531, latitude: 37.76155490343394)
-                guard let marker = currentLocationGem else {
-                    let marker = markerAdd()
-                    if marker == 0 { return } // Didn't initialize correctly.
-                    currentLocationGem = marker;
-                    markerSetPoint(marker, coordinates: point)
-                    //TODO: Update once scene updates are properly synchronous - { style: ux-location-gem-overlay, interactive: true, sprite: ux-current-location, size: 36px, collide: false }
-                    markerSetStyling(marker, styling: "{ style: 'points', color: 'white', size: [25px, 25px], order:500, collide: false }")
-                    animateToPosition(point, withDuration: 2.0)
-                    animateToZoomLevel(15, withDuration: 2.0)
-                    return
-                }
-                markerSetVisible(marker, visible: true)
-            } else {
-                guard let marker = currentLocationGem else {
-                    return;
-                }
-                markerSetVisible(marker, visible: false)
+    //! Returns whether or not the map was centered on the device's current location
+    public func centerOnCurrentLocation(shouldCenter: Bool) -> Bool {
+        guard let marker = currentLocationGem else { return false }
+        if marker == 0 { return false } // Invalid Marker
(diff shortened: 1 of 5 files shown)


ios-sdk/src/MapViewController.swift
```diff
@@ -0,0 +1,68 @@
...
+                markerSetVisible(marker, visible: true)
+                animateToPosition(point, withDuration: 2.0)
+                animateToZoomLevel(15, withDuration: 2.0)
+                currentLocationGem = marker
+            } else {
+                guard let marker = currentLocationGem else {
+                    // Not showing current location
+                    return;
+                }
+                markerRemove(marker)
```
karimnaaji: We probably want to play with the visibility of the marker here, since resources for the marker have been created already it would make a better usage of it.

ios-sdk/src/MapViewController.swift
```diff
@@ -0,0 +1,68 @@
...
+                    return;
+                }
+                print("Queuing scene update for current location")
+                let point = TGGeoPoint(longitude: -122.44880676269531, latitude: 37.76155490343394)
+                let marker = markerAdd()
+                //TODO: Update once scene updates are properly synchronous - { style: ux-location-gem-overlay, interactive: true, sprite: ux-current-location, size: 36px, collide: false }
+
+                markerSetStyling(marker, styling: "{ style: 'points', color: 'white', size: [25px, 25px], order:500, collide: false }")
+                markerSetPoint(marker, coordinates: point)
+                markerSetVisible(marker, visible: true)
```
karimnaaji: This call is superfluous when the marker has just been created.

msmollin: Gotcha. Didn't know markers were visible by default good to know.

ios-sdk/src/MapViewController.swift
msmollin: Probably not a bad idea. I can update that in the next PR.

ecgreb: Toggling visibility rather than adding/removing the marker seems like good practice for optimization of this and possibly other marker overlays to come.

If we're not going to update the behavior in this change set we should make an issue to update it later so we don't forget.

ios-sdk/src/MapViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+                guard let marker = currentLocationGem else {
+                    let marker = markerAdd()
+                    if marker == 0 { return } // Didn't initialize correctly.
+                    currentLocationGem = marker;
+                    markerSetPoint(marker, coordinates: point)
+                    //TODO: Update once scene updates are properly synchronous - { style: ux-location-gem-overlay, interactive: true, sprite: ux-current-location, size: 36px, collide: false }
+                    markerSetStyling(marker, styling: "{ style: 'points', color: 'white', size: [25px, 25px], order:500, collide: false }")
+                    animateToPosition(point, withDuration: 2.0)
+                    animateToZoomLevel(15, withDuration: 2.0)
+                    currentLocationGem = marker
```
msmollin: So it is! Fixing.

msmollin:
Agreed on the first point.

The second point is a little more difficult seeing as this is a setter/getter. I could switch it over to be a function but that changes my intent here of keeping it a more simple API. Let me think on that a bit.

msmollin:
@karimnaaji updated per your comments.

ios-sdk/src/MapViewController.swift
```diff
@@ -0,0 +1,49 @@
...
+import TangramMap
+
+public class MapViewController: TGMapViewController {
+
+    var currentLocationGem: TGMapMarkerId?
+    //TODO: This will eventually get refactored out to be whatever the current location is from CoreLocation
+    let tempPoint = TGGeoPoint(longitude: -122.44880676269531, latitude: 37.76155490343394)
+
+    //! Returns whether or not the map was centered on the device's current location
+    public func centerOnCurrentLocation(shouldCenter: Bool) -> Bool {
```
karimnaaji: what is `shouldCenter` used for?

msmollin: Leftover from testing different ways to center. Removed.

PR #95, opened 2016-12-02
Add CoreLocation updating for current location

msmollin:
Next step of #31

### DIFF
--- a/ios-sdk/src/LocationManager.swift
+++ b/ios-sdk/src/LocationManager.swift
@@ -40,7 +40,7 @@ public class LocationManager: NSObject, CLLocationManagerDelegate {
 
     public func refreshCurrentLocation() -> CLLocation? {
         if CLLocationManager.locationServicesEnabled() &&
-           ( CLLocationManager.authorizationStatus() == .AuthorizedAlways || CLLocationManager.authorizationStatus() == .AuthorizedWhenInUse ) {
+           (CLLocationManager.authorizationStatus() == .AuthorizedAlways || CLLocationManager.authorizationStatus() == .AuthorizedWhenInUse) {
             currentLocation = coreLocationManager.location
             return coreLocationManager.location
         }
--- a/ios-sdk/src/LocationManager.swift
+++ b/ios-sdk/src/LocationManager.swift
@@ -74,7 +74,7 @@ public class LocationManager: NSObject, CLLocationManagerDelegate {
     }
 
     public func locationManager(manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
-        if let location = locations.first {
+        if let location = locations.last {
             delegate?.locationDidUpdate?(location)
         }
     }
--- a/ios-sdk/src/LocationManager.swift
+++ b/ios-sdk/src/LocationManager.swift
@@ -1,0 +1,82 @@
+//
+//  LocationManager.swift
+//  ios-sdk
+//
+//  Created by Matt Smollinger on 11/22/16.
+//  Copyright © 2016 Mapzen. All rights reserved.
+//
+
+import UIKit
+import CoreLocation
+
+@objc public protocol LocationManagerDelegate {
+    optional func authorizationDidSucceed()
+    optional func authorizationDenied()
+    optional func authorizationRestricted()
+    optional func locationDidUpdate(location: CLLocation)
+}
+
+
+public class LocationManager: NSObject, CLLocationManagerDelegate {
+
+    public static let sharedManager = LocationManager()
+    public var currentLocation: CLLocation?
+    public weak var delegate: LocationManagerDelegate?
+
+    private let coreLocationManager = CLLocationManager()
+    private override init(){
+        super.init()
+        coreLocationManager.delegate = self
+    }
+
+    public func requestAlwaysAuthorization() {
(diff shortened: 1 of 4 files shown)


ios-sdk/src/MapViewController.swift
```diff
@@ -9,18 +9,20 @@
...
 
     var currentLocationGem: TGMapMarkerId?
     //TODO: This will eventually get refactored out to be whatever the current location is from CoreLocation
-    let tempPoint = TGGeoPoint(longitude: -122.44880676269531, latitude: 37.76155490343394)
+    var lastSetPoint: TGGeoPoint?
 
     //! Returns whether or not the map was centered on the device's current location
     public func centerOnCurrentLocation() -> Bool {
         guard let marker = currentLocationGem else { return false }
+        guard let point = lastSetPoint else { return false }
```
karimnaaji: 👍

ios-sdk/src/MapViewController.swift
```diff
@@ -9,18 +9,20 @@
...
         if marker == 0 { return false } // Invalid Marker
-        animateToPosition(tempPoint, withDuration: 2.0)
+        animateToPosition(point, withDuration: 2.0)
```
karimnaaji: Might be good to pass those as a parameter to `centerOnCurrentLocation`

ios-sdk/src/MapViewController.swift
```diff
@@ -46,4 +48,23 @@ public class MapViewController: TGMapViewController {
...
+    }
+
+    //MARK: - LocationManagerDelegate
+
+    public func locationDidUpdate(location: CLLocation) {
+        guard let marker = currentLocationGem else {
+            return
+        }
+        lastSetPoint = TGGeoPoint(longitude: location.coordinate.longitude, latitude: location.coordinate.latitude)
+        markerSetPoint(marker, coordinates: TGGeoPoint(longitude: location.coordinate.longitude, latitude: location.coordinate.latitude))
```
karimnaaji: Can use `lastSetPoint`.

ios-sdk/src/LocationManager.swift
```diff
@@ -0,0 +1,82 @@
...
+            delegate?.authorizationDenied?()
+        case .Restricted:
+            delegate?.authorizationRestricted?()
+        default:
+            print("Not Authorized")
+        }
+    }
+
+    public func locationManager(manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
+        if let location = locations.first {
```
karimnaaji: From: https://developer.apple.com/reference/corelocation/cllocationmanagerdelegate/1423615-locationmanager

>Therefore, the most recent location update is at the end of the array.

Seems that what you want is the last one?

ios-sdk/src/MapViewController.swift
msmollin: You can but this is a more succinct API because `lastSetPoint` is an optional, and so i'd have to unwrap `lastSetPoint` in some fashion, either using `if let` / `guard let` or force unwrapping. It's a bit of an annoyance with swift.

ios-sdk/src/MapViewController.swift
msmollin: Brilliant (and obvious) good suggestion. Will do.

ios-sdk/src/MapViewController.swift
karimnaaji: Oh I see.

ios-sdk/src/LocationManager.swift
msmollin: Innnnnteresting. The headers for `CLLocationManagerDelegate` don't state the order and so I assumed the first one in the array. Will update.

```     *  Discussion:
     *    Invoked when new locations are available.  Required for delivery of
     *    deferred locations.  If implemented, updates will
     *    not be delivered to locationManager:didUpdateToLocation:fromLocation:
     *
     *    locations is an array of CLLocation objects in chronological order.
```

ios-sdk/src/LocationManager.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    }
+
+    public func requestWhenInUseAuthorization() {
+        coreLocationManager.requestWhenInUseAuthorization()
+    }
+
+    public func refreshCurrentLocation() -> CLLocation? {
+        if CLLocationManager.locationServicesEnabled() &&
+           ( CLLocationManager.authorizationStatus() == .AuthorizedAlways || CLLocationManager.authorizationStatus() == .AuthorizedWhenInUse ) {
```
karimnaaji: Extra space

ecgreb:
@karimnaaji this use case should be handled by the client app developer. `MapViewController.showCurrentLocation()` should not be called if the user denies location access. If it is still invoked when location permissions have not been granted then it should do nothing.

ios-sdk/src/LocationManager.swift
msmollin: Fixed

ios-sdk/src/MapViewController.swift
```diff
@@ -9,41 +9,61 @@
...
+        if (shouldShowCurrentLocation){
```
msmollin: Fixed

karimnaaji:
Last changes looks good 👍

PR #113, opened 2016-12-13
Restructure project and remove unused files

### LINKED ISSUE #109: SDK and sample app in separate top level folders
Currently the SDK classes are in a subfolder of the demo app.

https://github.com/mapzen/ios/tree/master/ios-sdk/src

To make the project easier to navigate the SDK and sample app classes should each be in separate top level folders named accordingly.

msmollin:
Cleans up unused font files and fixes #109 

Also updates the Podfile.lock to grab the latest tangram-es framework


ecgreb:
Looking nice and tidy!

What is the size difference in the SDK (and demo app binary) after removing the unused font files?

msmollin:
Approximately 9.4 MB of files were removed, and I think that directly translates to the sample app. I think only 7.2 MB of them were actually getting imported into the SDK though, but still a very meaningful number.

PR #127, opened 2016-12-21
Implement "master switch" for Location layer, unit tests, and general refactoring.

### LINKED ISSUE #31: Implement find me button to recenter the map

msmollin:
Fixes #31


ios-sdkTests/MapViewControllerTests.swift
```diff
@@ -0,0 +1,57 @@
...
+
+        //Now for flipping it back to false
+        controller.showFindMeButon(false)
+
+        //Tests
+        XCTAssert(controller.findMeButton.hidden == true)
+        XCTAssert(controller.findMeButton.enabled == false)
+    }
+
+    func testDisableLocation() {
```
ecgreb: Should there also be a `testEnableLocationLayer()`?

src/MapViewController.swift
```diff
@@ -54,8 +54,16 @@ public class MapViewController: TGMapViewController, LocationManagerDelegate, TG
         return true
     }
 
+    public func enableLocationLayer(enabled: Bool) {
+        showCurrentLocation(enabled)
+        showFindMeButon(enabled)
+        enabled ? LocationManager.sharedManager.startUpdatingLocation() : LocationManager.sharedManager.stopUpdatingLocation()
+        shouldFollowCurrentLocation = enabled
```
ecgreb: `shouldFollowCurrentLocation` is already set as part of call to `showCurrentLocation(enabled)` no need to set again here.

ecgreb: Nvm I got `shouldShowCurrentLocation` mixed up with `shouldFollowCurrentLocation`

ios-sdkTests/MapViewControllerTests.swift
msmollin: Why yes, we should. I'll do that :)

ecgreb: 👍

msmollin:
@matteblair / @karimnaaji  could use your input of the use of `viewDidLoad` inside the test harness here: https://github.com/mapzen/ios/pull/127/files#diff-29b94e593e9465be78ea4774e6e81253R93

Basically I'm doing that to force TGMapViewController to call `setupGL` and load a `TangramMap` instance. So, two questions:

1) Is that OK to do without having actually added the instance of TestMapViewController that's under test to the view hierarchy?
2) What side effects does that induce that could be detrimental to unit testing?

My first thought for the second point is that it could start loading tiles theoretically, which would be a bad thing. My thought is to subclass `TGHTTPHandler` and override the `downloadRequestAsync` method to not actually start a request. Will that keep Tangram off the network then?

msmollin:
Hmmm, so, some further musings: the reason I'm trying to get a `TangramMap` instance is to be able to get Tangram to vend out a map marker ID. If the iOS SDK doesn't have a valid map marker ID, it won't process some of the logic I'd like to have under test.

Looking at the `markerSetVisible` and `markerSetPoint` commands in `TGMapViewController`, it would appear that without an instance of `TangramMap`, the view controller will simply ignore those calls and return NO, which is fine for the iOS SDK (for now anyway). So, I could just expose the map marker ID in the code, pass in a mock, and call it done. That avoids having to spin up Tangram and potentially get networking calls going.

However, I could potentially see this approach not working everywhere. Thoughts?

karimnaaji:
@msmollin if you're doing tests that are only relative to the SDK, is it possible to have a mock for the MapViewController of tangram? In your case you create the method for markers that returns a valid id. I would not try to unit test something that creates an opengl context and could do network requests. 

If you setup the scene for the controller, network requests would potentially start depending on what method calls happen on the controller, so overriding the HttpHandler is a good idea! I would explore how a mock would work for those classes.

msmollin:
Ah yeah you're right. I could override markerAdd and have it always return a valid marker ID. I'm already doing something akin here: https://github.com/mapzen/ios/pull/127/files#diff-29b94e593e9465be78ea4774e6e81253R14 just to make it a little easier to test some of the internal state machine.

I'll give that a try. Thanks.

msmollin:
@karimnaaji / @ecgreb updated per @karimnaaji's suggestions. Sped up the tests a bit too! Nothing huge: 0.05s -> 0.011s in a few of them, but it was nice to see as a bonus.

PR #130, opened 2016-12-22
Fix crash related to selectors and static functions not mixing

msmollin:


src/MapViewController.swift
```diff
@@ -17,7 +17,21 @@ public class MapViewController: TGMapViewController, LocationManagerDelegate, TG
...
+    public var findMeButton = UIButton(type: .Custom)
+
+    init(){
+        super.init(nibName: nil, bundle: nil)
+        defer {
+            findMeButton = createFindMeButton()
+        }
+    }
+    
+    required public init?(coder aDecoder: NSCoder) {
```
msmollin: This is needed because Apple marked `init?(coder:)` as required implementation anytime you add a convenience initializer like `init()`. No idea why but it's been a thing since iOS 9.

src/MapViewController.swift
```diff
@@ -17,7 +17,21 @@ public class MapViewController: TGMapViewController, LocationManagerDelegate, TG
     var lastSetPoint: TGGeoPoint?
     var shouldShowCurrentLocation = false
     public var shouldFollowCurrentLocation = false
-    public var findMeButton = MapViewController.createFindMeButton()
+    public var findMeButton = UIButton(type: .Custom)
```
ecgreb: Does `findMeButton` need to be initialized here when its going to be set by `createFindMeButton()` in the constructor?

msmollin: Just to clarify - answer is yes cuz Swift. There may be a cleaner way of doing it though. I'll look into it for future cleanup.

PR #134, opened 2017-01-02
Update podspec to fix import issues and some linting warnings

msmollin:
This still won't fully fix imports until we submit on the road and tangram to cocoapods trunk, but at least the imports work now.


ios-sdk.podspec
```diff
@@ -5,15 +5,12 @@ Pod::Spec.new do |s|
 
   s.summary           = 'Mapzen iOS SDK'
   s.description       = 'Mapzen iOS SDK'
-  s.homepage          = 'https://mapzen.com/products/ios/'
+  s.homepage          = 'https://mapzen.com/projects/mobile/'
```
ecgreb: This url is likely to change again in the near future but this is the best one to use for now. Alternatively you could use https://mapzen.com/documentation/ios/.

PR #135, opened 2017-01-04
Reformat most files to two space indent, and small additions

### LINKED ISSUE #132: Sample app should use current location of user for routing start point
Since, you know, we have nice things now like requesting the current location of the user.

msmollin:
Adds a couple small functions to location manager, and cleans up a couple
minor bugs.

Fixes #132

### DIFF
--- a/SampleApp/AutocompleteTableVC.swift
+++ b/SampleApp/AutocompleteTableVC.swift
@@ -11,11 +11,11 @@ import MapKit
 import CoreLocation
 import Pelias
 
-class AutocompleteTableVC: UITableViewController, UISearchResultsUpdating, UISearchBarDelegate, CLLocationManagerDelegate {
+class AutocompleteTableVC: UITableViewController, UISearchResultsUpdating, UISearchBarDelegate, LocationManagerDelegate {
   
   let searchController = UISearchController(searchResultsController: nil)
   var results: [PeliasMapkitAnnotation]?
-  let manager = CLLocationManager()
+  let manager = LocationManager.sharedManager
   var currentLocation: CLLocation?
   
   override func viewDidLoad() {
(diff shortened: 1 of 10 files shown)


SampleApp/AutocompleteTableVC.swift
```diff
@@ -27,21 +27,19 @@ class AutocompleteTableVC: UITableViewController, UISearchResultsUpdating, UISea
     self.definesPresentationContext = true
     searchController.searchBar.sizeToFit()
     manager.delegate = self
-    PeliasSearchManager.sharedInstance.urlQueryItems = [NSURLQueryItem(name: "apiKey", value: "mapzen-2qQR7SX")]
+    PeliasSearchManager.sharedInstance.urlQueryItems = [NSURLQueryItem(name: "api_key", value: "mapzen-2qQR7SX")]
```
ecgreb: We should create an issue to remove hard-coded API keys prior to launch.

The sample app in the Android SDK has developers provide their own Mapzen API key via an [xml resource file](https://github.com/mapzen/android/blob/master/samples/mapzen-android-sdk-sample/src/main/res/values/mapzen.xml) or [Gradle properties](https://github.com/mapzen/android/blob/master/samples/mapzen-android-sdk-sample/build.gradle#L4).

What we the best approach on iOS? A `.plist` file?

msmollin: Plist or code-based is probably best. Now that we have a decent way todo it in Tangram-es we can have either/both.

SampleApp/RoutingViewController.swift
```diff
@@ -41,11 +41,23 @@ class RoutingViewController: UIViewController, RoutingSearchDelegate {
     }
   }
 
+  override func viewDidAppear(animated: Bool) {
+    super.viewDidAppear(animated)
+    if LocationManager.sharedManager.isAlwaysAuthorized() || LocationManager.sharedManager.isInUseAuthorized() {
+      LocationManager.sharedManager.startUpdatingLocation()
```
ecgreb: For the routing example do we need constant location updates or would it suffice to request a single location when the request is made?

SampleApp/AutocompleteTableVC.swift
ecgreb: I like providing both options in the SDK if possible.

SampleApp/RoutingViewController.swift
msmollin: I went this way because it's the way I would except people to implement it in an app, if that makes sense? 

The problem with requesting location in a "one off" scenario is the location may not be current, or core location / our location manager may not have one at all, so we would have to heat up the modem/GPS to get a fix and that could take a bit, meanwhile the user is waiting. Doing it this way, the `LocationManager`'s `.currentLocation` property is getting updated and caching the last known good location, which we're then using on  [line 58](https://github.com/mapzen/ios/pull/135/files/6a963c975a01eb8b4225fe22cee38f14776d2267#diff-5e8f16b3beb0c57a7af33e8a0c0c53b8R58) when someone makes a request to route.

Obviously it's a little overkill for a sample app but I think it provides the best UX.

ecgreb: 👍

SampleApp/AutocompleteTableVC.swift
msmollin: https://github.com/mapzen/ios/issues/136

Issue opened to track.

Issue #136, opened 2017-01-04
Update SDK to add central api_key insertion layer, and update sample app to use


msmollin:
Sample app can load the API KEY from a plist file or whatever makes sense.

PR #138, opened 2017-01-09
Add search pins to the demo app & annotation convenience methods

### LINKED ISSUE #69: Add search result pins to demo app
Split screen showing map with search pins and list view of all results

msmollin:
Fixes #69


SampleApp/SearchPinsViewController.swift
```diff
@@ -0,0 +1,42 @@
...
+    textField.resignFirstResponder()
+    return false
+  }
+
+  func textFieldDidEndEditing(textField: UITextField) {
+    let geopoint = GeoPoint(location: LocationManager.sharedManager.currentLocation)
+    let weakSelf = self
+    var searchConfig = PeliasSearchConfig(searchText: textField.text!) { (response) in
+      guard let newAnnotations = response.parsedMapItems() else { return }
+      let strongSelf = weakSelf
```
karimnaaji: Is there any syntax in the language that permits that?

src/MapViewController.swift
```diff
@@ -74,6 +76,35 @@ public class MapViewController: TGMapViewController, LocationManagerDelegate, TG
...
+      markerSetPoint(newMarker, coordinates: TGGeoPoint(coordinate: annotation.coordinate))
+      markerSetStyling(newMarker, styling: "{ style: sdk-point-overlay, sprite: ux-search-active, size: [24, 36px], collide: false }")
+      currentAnnotations[annotation] = newMarker
+
+    }
+  }
+
+  public func remove(annotation: PeliasMapkitAnnotation) {
+    guard let markerId = currentAnnotations[annotation] else { return }
+    markerRemove(markerId)
```
karimnaaji: Just in case Tangram fails it would be good to check the return status of `markerRemove` before, I also suggests to return a boolean that gets this status from this function.

src/MapViewController.swift
```diff
@@ -74,6 +76,35 @@ public class MapViewController: TGMapViewController, LocationManagerDelegate, TG
...
 
+  public func loadScene(named: String, apiKey: String) {
+    self.loadSceneFile(named)
+    self.queueSceneUpdate("sources.mapzen.url_params", withValue: "{ api_key: \(apiKey)}")
+    self.applySceneUpdates()
+  }
+
+  public func add(annotations: [PeliasMapkitAnnotation]){
+    for annotation in annotations {
+      let newMarker = self.markerAdd()
```
karimnaaji: `newMarker` would need to be checked against 0 for its validity.

src/MapViewController.swift
```diff
@@ -74,6 +76,35 @@ public class MapViewController: TGMapViewController, LocationManagerDelegate, TG
...
+
+  public func remove(annotation: PeliasMapkitAnnotation) {
+    guard let markerId = currentAnnotations[annotation] else { return }
+    markerRemove(markerId)
+    currentAnnotations.removeValueForKey(annotation)
+  }
+
+  public func removeAnnotations() {
+    for (annotation, markerId) in currentAnnotations {
+      markerRemove(markerId)
```
karimnaaji: Same here, we need to check return status incoming from tangram.

src/PeliasMapkitExtensions.swift
```diff
@@ -144,3 +144,10 @@ extension SearchBoundaryRect {
     self.minLatLong = GeoPoint(latitude: maxCoordinate.latitude, longitude: minCoordinate.longitude)
   }
 }
+
+public extension GeoPoint {
```
karimnaaji: Do you think there is any way we could reduce the number of structures made to represent a location? Mostly thinking of `TGGeoPoint` and this one.

SampleApp/SearchPinsViewController.swift
```diff
@@ -0,0 +1,42 @@
...
+import UIKit
+import Pelias
+
+class SearchPinsViewController: MapViewController, UITextFieldDelegate {
+
+  @IBOutlet weak var searchField: UITextField!
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    self.loadScene("scene.yaml", apiKey: "mapzen-2qQR7SX")
```
ecgreb: Good idea. We should create an issue to centralize API key management in the SDK (if it does not already exits).

I would be in favor of adding that issue to the beta milestone as its not really required to get early feedback on the core functionality of the SDK which is our main goal for alpha.

src/PeliasMapkitExtensions.swift
ecgreb: Not sure about the specifics on extensions but I think the issue here is that Pelias as a standalone dependency doesn't know about Tangram classes right?

SampleApp/SearchPinsViewController.swift
msmollin: https://github.com/mapzen/ios/issues/136

SampleApp/SearchPinsViewController.swift
msmollin: Permits what?

src/PeliasMapkitExtensions.swift
msmollin: @ecgreb is correct, and Pelias is built to not require CoreLocation (same for Tangram-es) so we need a structure that allows for that.

One of the plans I have post-1.0 is to build a "Mapzen Utilities" library that all 3 of these would depend upon (cuz On the Road has its own version as well) and would utilize.

msmollin: Opened an issue to track: https://github.com/mapzen/ios/issues/141

Feel free to post additional thoughts there.

SampleApp/SearchPinsViewController.swift
karimnaaji: Permits to not store a local reference to `self` and directly give it to a lambda function as a parameter of the handler.

msmollin: I wish :( It's got the same issue as Objective-C in that regard.

Although I just realized I didn't actually define `weakSelf` as `weak` so this actually isn't doing what I expect it to haha. I'll fix that.

SampleApp/SearchPinsViewController.swift
karimnaaji: Sorry I haven't seen that this has already been discussed in https://github.com/mapzen/ios/pull/135.

msmollin:
@karimnaaji / @ecgreb this has been updated with necessary checks. I've disabled the unit tests for those pieces until we can move around how Tangram-es handles it's map initialization.

ecgreb:
Is there an open issue for updating map initialization so we can uncomment these tests?

ecgreb:
Nvm found it! https://github.com/tangrams/tangram-es/issues/1220

msmollin:
Yep, I made a linked issue here for us to track in iOS - https://github.com/mapzen/ios/issues/142

PR #144, opened 2017-01-16
Centralize API Key Manager and Implement Route Line Drawing

### LINKED ISSUE #70: Add route line to demo app
Split screen demo with map plus route line and direction list

### LINKED ISSUE #136: Update SDK to add central api_key insertion layer, and update sample app to use

msmollin:
I got tired of copy-pasting API keys all over the place so I introduced the functionality to centralize the API key management. This fixes #136.

Then I got onto the original reason for this PR - route line drawing. This adds some convenience layer functionality to integrate OTR with the map view. Fixes #70


SampleApp/AppDelegate.swift
```diff
@@ -16,6 +16,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
   func application(application: UIApplication, didFinishLaunchingWithOptions launchOptions: [NSObject: AnyObject]?) -> Bool {
     // Override point for customization after application launch.
+    MapzenManager.sharedManager.apiKey = "mapzen-2qQR7SX"
```
ecgreb: Its great to have this centralized in a single location now. Before we release the final version of the SDK this API key should be replaced with something like `[YOUR_MAPZEN_API_KEY]` and instructions should be added to the README for setting the API key. This will help prevent accidental re-use of house demo API keys by external developers.

https://github.com/mapzen/ios/issues/145

SampleApp/SearchPinsViewController.swift
```diff
@@ -15,7 +15,7 @@ class SearchPinsViewController: MapViewController, UITextFieldDelegate {
 
   override func viewDidLoad() {
     super.viewDidLoad()
-    self.loadScene("scene.yaml", apiKey: "mapzen-2qQR7SX")
+    let _ = try? loadScene("scene.yaml")
     PeliasSearchManager.sharedInstance.urlQueryItems = [NSURLQueryItem(name: "api_key", value: "mapzen-2qQR7SX")]
```
ecgreb: Can we update/wrap the search manager to also pull the API key from `MapzenManager`?

ios-sdkTests/MapzenManagerTests.swift
```diff
@@ -0,0 +1,87 @@
...
+
+  func testApiKeyMapNotSet() {
+    //Setup
+    MapzenManager.sharedManager.apiKey = nil
+    let map = MapViewController()
+
+    //Tests
+    XCTAssertThrowsError(try map.loadScene("someYamlThing.yaml", apiKey: nil)) { (error) -> Void in
+      let error = error as NSError
+      XCTAssertTrue(error.code == MZError.APIKeyNotSet.rawValue)
```
ecgreb: 👍

SampleApp/SearchPinsViewController.swift
ecgreb: Hmm after rereading the tests maybe it is and this line can be removed?

src/MapViewController.swift
```diff
@@ -128,6 +141,49 @@ public class MapViewController: TGMapViewController, LocationManagerDelegate, TG
...
+    //TODO: We eventually should support N number of routes.
+    if let routeMarker = currentRouteMarker {
+      //We don't throw if the remove fails here because we want to silently replace the route
+      currentRouteMarker = nil
+      markerRemove(routeMarker)
+    }
+    let routeLeg = route.legs[0]
+    let polyLine = TGGeoPolyline(size: UInt32(routeLeg.coordinateCount))
+
+    //TODO: Need to investigate more if this is a bug in OTR or if valhalla returns null island at the end of their requests?
```
ecgreb: I have never seen 0,0 returned as the final set of coordinates returned by the Valhalla service. My guess would be something in OTR is appending this as the final value. Where is the decoding of the polyline implemented?

msmollin: Yeah that's what I'm figuring. Here's the decode implementation:
https://github.com/mapzen/on-the-road_ios/blob/master/OTRRouting/OTRRoutingResultLeg.m#L53

src/MapViewController.swift
```diff
@@ -128,6 +141,49 @@ public class MapViewController: TGMapViewController, LocationManagerDelegate, TG
...
+      print("Next Point: \(point)")
+      polyLine.addPoint(TGGeoPoint(coordinate: point))
+    }
+    let marker = markerAdd()
+    if marker == 0 {
+      throw NSError(domain: MapViewController.MapzenGeneralErrorDomain,
+                    code: MZError.GeneralError.rawValue,
+                    userInfo: nil)
+    }
+    markerSetStyling(marker, styling: "{ style: lines, color: blue, width: 10px, order: 5000 }")
```
msmollin: Gotcha. Will update.

SampleApp/SearchPinsViewController.swift
msmollin: Ewps. Yeah it is. Good catch. Will remove.

SampleApp/AppDelegate.swift
msmollin: 👍

src/MapViewController.swift
```diff
@@ -128,6 +141,49 @@ public class MapViewController: TGMapViewController, LocationManagerDelegate, TG
...
   }
 
+  public func display(route: OTRRoutingResult) throws {
+    //TODO: We eventually should support N number of routes.
+    if let routeMarker = currentRouteMarker {
+      //We don't throw if the remove fails here because we want to silently replace the route
+      currentRouteMarker = nil
+      markerRemove(routeMarker)
+    }
+    let routeLeg = route.legs[0]
```
karimnaaji: What is the significance of the other indices? We just don't need them?

msmollin: Ah, good catch. This is something I meant to address when I was moving stuff into the SDK from the sample app.

Valhalla has a concept of "trips", which can have multiple "legs". Each leg is analogous to a single poly-line in Tangram-es. I actually don't know if we support multi-leg routing in Android... @ecgreb / @sarahlensing might have a better idea?

msmollin: My thought is we could support it rather easily with a single poly line (just adding up the sizes of the individual legs) and then the implementing dev could just drop extra markers where they know there are stops involved, rather than having multiple polylines running around to keep track of.

ecgreb: Multiple routing waypoints is not yet supported in the Android SDK

msmollin: So do we want to support it here as a single polyline? Or just punt on it for now.

ecgreb: You can see here in fact on the road Android also only cares about `leg[0]` currently.

https://github.com/mapzen/on-the-road_android/blob/master/library/src/main/java/com/mapzen/valhalla/Route.kt#L69-L70

Let's punt for now and make an issue to add full support for waypoints later.

PR #153, opened 2017-01-27
Update podspec to 0.1.0 for release

msmollin:
Ideally given how the podspec dependencies are defined we shouldn't have to alter this podspec further as we update the beta releases of the underlying frameworks.

### DIFF
--- a/ios-sdk.podspec
+++ b/ios-sdk.podspec
@@ -1,27 +1,27 @@
 Pod::Spec.new do |s|
 
-  s.name    = 'ios-sdk'
+  s.name    = 'Mapzen-ios-sdk'
   s.version = '0.1.0'
 
   s.summary           = 'Mapzen iOS SDK'
-  s.description       = 'Mapzen iOS SDK'
+  s.description       = 'The Mapzen iOS SDK is a thin wrapper that packages up everything you need to use Mapzen services in your iOS applications. It also simplifies setup, installation, API key management, and generally makes your life better.'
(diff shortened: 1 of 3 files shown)


ios-sdk.podspec
```diff
@@ -1,27 +1,27 @@
...
-  s.description       = 'Mapzen iOS SDK'
+  s.description       = 'The Mapzen iOS SDK is a thin wrapper that packages up everything you need to use Mapzen services in your iOS applications. It also simplifies setup, installation, API key management, and generally makes your life better.'
   s.homepage          = 'https://mapzen.com/projects/mobile/'
-  s.license           = { :type => 'MIT', :file => 'LICENSE.md' }
+  s.license           = { :type => 'Apache License, Version 2.0', :file => 'LICENSE.md' }
   s.author            = { 'Mapzen' => 'ios-support@mapzen.com' }
   s.social_media_url  = 'https://twitter.com/mapzen'
   s.documentation_url = 'https://mapzen.com/documentation/ios/'
-  s.source           =  { :git => 'https://github.com/mapzen/ios.git', :branch => 'master' }
+  s.source           =  { :git => 'https://github.com/mapzen/ios.git', :tag => "v#{s.version}" }
```
ecgreb: We will need to manually push the tag `v0.1.0` to this repo after this merges, correct?

msmollin: Yeah. I was doing some reading and it seems like if you tag on the branch and do a merge commit it is fine, but tags don't really live in branches so it's weird.

PR #154, opened 2017-01-31
update syntax

kkowalsky:
hey @msmollin is this correct? specifically at line 13 does it need to be in single quotes


msmollin:
Yep that's me being sloppy :) Approved.

PR #167, opened 2017-02-15
Wrap TGMapViewController

sarahlensing:
This PR updates the SDK's main map entry point `MapViewController` so that it wraps `TGMapViewController`

- `MapViewController` now creates an instance of `TGMapViewController` and adds its view as a subview
- Most of the `TGMapViewController` methods have been added to `MapViewController`s interface and forward calls to the internal `TGMapViewController`
- Methods that have been excluded from `MapViewController` include:
```
public func pickFeatureAt(screenPosition: CGPoint)
public func pickLabelAt(screenPosition: CGPoint)
public func pickMarkerAt(screenPosition: CGPoint)
public func setPickRadius(logicalPixels: Float)
public func requestRender()
public func renderOnce()
public func update()
```
- Adds tests to `MapViewControllerTest` for these changes

A future PR will:
- Add wrappers for `TGMapViewDelegate` and `TGRecognizerDelegate` as well as set an internal delegate in `MapViewController` to handle forwarding calls to the external delegates
- Transparently add back support for these excluded methods:
```
public func pickFeatureAt(screenPosition: CGPoint)
public func pickLabelAt(screenPosition: CGPoint)
public func pickMarkerAt(screenPosition: CGPoint)
```
(when a delegate is set on `MapViewController`, we can have an internal delegate handle picking features, labels, markers)

### DIFF
--- a/src/MapViewController.swift
+++ b/src/MapViewController.swift
@@ -31,6 +31,51 @@ public class MapViewController: UIViewController, LocationManagerDelegate, TGRec
   public var findMeButton = UIButton(type: .Custom)
   public var currentAnnotations: [PeliasMapkitAnnotation : TGMapMarkerId] = Dictionary()
 
+  public var cameraType: TGCameraType {
+    set {
+      tgViewController.cameraType = cameraType
+    }
+    get {
+      return tgViewController.cameraType
+    }
+  }
+  
+  public var position: TGGeoPoint {
+    set {
+      tgViewController.position = position
+    }
+    get {
+      return tgViewController.position
+    }
+  }
+  
+  public var zoom: Float {
+    set {
+      tgViewController.zoom = zoom
+    }
+    get {
+      return tgViewController.zoom
+    }
+  }
+  
+  public var rotation: Float {
+    set {
+      tgViewController.rotation = rotation
+    }
+    get {
+      return tgViewController.rotation
+    }
+  }
+  
+  public var tilt: Float {
+    set {
+      tgViewController.tilt = tilt
+    }
+    get {
+      return tgViewController.tilt
+    }
+  }
+
   init(){
     super.init(nibName: nil, bundle: nil)
     defer {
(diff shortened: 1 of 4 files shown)


src/MapViewController.swift
```diff
@@ -17,12 +17,12 @@ import OnTheRoad
...
 
-public class MapViewController: TGMapViewController, LocationManagerDelegate, TGRecognizerDelegate {
+public class MapViewController: UIViewController, LocationManagerDelegate, TGRecognizerDelegate {
 
   //Error Domains for NSError Appeasement
   public static let MapzenGeneralErrorDomain = "MapzenGeneralErrorDomain"
-
-
+  
+  var tgViewController: TGMapViewController = TGMapViewController()
```
karimnaaji: I might be missing something but isn't that the intent of this PR? If we mark it as public it might be creating ambiguous access to the method of the Tangram controller, since someone would have two ways to interact the map view, through this accessor and through the wrapped methods.

tallytalwar: Is it similar to the use case here: https://github.com/mapzen/android/blob/master/core/src/main/java/com/mapzen/android/graphics/MapView.java#L106

Where one gives access to the underlying TangramMap object, but still the recommended way is to use the MapView (wrapper methods).

msmollin: Yeah what @tallytalwar said.

sarahlensing: Fixed!

tallytalwar: Even though I pointed out the usage of the similar architecture design in android sdk. I still do not know a use case of when a developer will be using an accessor to the underlying TangramMap! Will be good to know of one :D.

ecgreb: On Android `MapView` and `MapController` are separate objects whereas in iOS `TGMapViewController` embodies both of these classes in that it is used to load the map _and_ provide controls like position, zoom, and the marker interface.

There are few cases where access to the underlying `MapView` would be needed on Android so I could see changing this to be protected in that developers could subclass `MapView` if they really needed access.

Exposing the `MapController` on Android and `TGMapViewController` on iOS makes more sense from the perspective that a developer may want to customize the behavior of the map in certain situations or have access to functionality that is not currently supported by the SDK.

PR #168, opened 2017-02-15
Add xcscmblueprint files to gitignore

sarahlensing:
- Updates `.gitignore` to remove `xcscmblueprint` files
- Removes workspace xcscmblueprint file


msmollin:
What's your reasoning behind removing this? It's attached to the shared scheme and tracks what repositories are in use in the workspace. I don't generally like removing items that attached to the shared scheme of a workspace.

sarahlensing:
I wanted to avoid a diff war for what seemed like negligible individual choices. I chose to checkout this repo without renaming the directory so it created this diff for the blueprint:
```
-    "BBDA5A20AD8873BE33564783C9AF146518693944" : "ios-sdk\/",
+    "BBDA5A20AD8873BE33564783C9AF146518693944" : "ios\/",
```
That combined with seeing it listed [here](https://github.com/github/gitignore/blob/master/Swift.gitignore) made me think it was ok to remove.

I don't have too much experience with this file so maybe this isn't a good enough reason to remove it!

msmollin:
Yeah I saw github added it, and I'm coming at it from a similar but opposing viewpoint: I don't have too much experience with it either so I am loathe to remove something apple thinks should be in a shared scheme, haha!

That said, less unnecessary churn I think is better, so lets go with it and if it breaks something we can bring it back. The beauty of source control!

PR #179, opened 2017-02-23
Add selectable marker example

### LINKED ISSUE #146: Demo app should have selectable markers
Make search result pins selectable and show more info about place when selected

sarahlensing:
- Updates the search demo to have selectable markers which show an alert when clicked
- Changes `MapFeatureSelectDelegate`, `MapLabelSelectDelegate`, and `MapMarkerSelectDelegate` to have non optional parameter

Closes #146

### DIFF
--- a/src/MapViewController.swift
+++ b/src/MapViewController.swift
@@ -49,15 +49,15 @@ public protocol MapShoveGestureDelegate : class {
 }
 
 public protocol MapFeatureSelectDelegate : class {
-  func mapController(_ controller: MapViewController, didSelectFeature feature: [AnyHashable : Any]?, atScreenPosition position: CGPoint)
+  func mapController(_ controller: MapViewController, didSelectFeature feature: [AnyHashable : Any], atScreenPosition position: CGPoint)
 }
 
 public protocol MapLabelSelectDelegate : class {
-  func mapController(_ controller: MapViewController, didSelectLabel labelPickResult: TGLabelPickResult?, atScreenPosition position: CGPoint)
+  func mapController(_ controller: MapViewController, didSelectLabel labelPickResult: TGLabelPickResult, atScreenPosition position: CGPoint)
 }
 
 public protocol MapMarkerSelectDelegate : class {
-  func mapController(_ controller: MapViewController, didSelectMarker markerPickResult: TGMarkerPickResult?, atScreenPosition position: CGPoint)
+  func mapController(_ controller: MapViewController, didSelectMarker markerPickResult: TGMarkerPickResult, atScreenPosition position: CGPoint)
 }
 
 public protocol MapTileLoadDelegate : class {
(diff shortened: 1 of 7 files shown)


src/MapViewController.swift
```diff
@@ -504,17 +507,17 @@ extension MapViewController : TGMapViewDelegate, TGRecognizerDelegate {
   
   open func mapView(_ mapView: TGMapViewController, didSelectFeature feature: [AnyHashable : Any]?, atScreenPosition position: CGPoint) {
     guard (feature != nil) else { return }
-    featureSelectDelegate?.mapController(self, didSelectFeature: feature, atScreenPosition: position)
+    featureSelectDelegate?.mapController(self, didSelectFeature: feature!, atScreenPosition: position)
```
msmollin: Nitpick: this will work but we can do better! 

Change the above guard statement to be:
`guard let feature = feature else { return }`

And then you can use the shadowed unwrapped optional in the call to the delegate, and thus avoid force unwrapping :)

src/MapViewController.swift
```diff
@@ -76,7 +76,8 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
   var currentRouteMarker: TGMapMarkerId?
   open var shouldFollowCurrentLocation = false
   open var findMeButton = UIButton(type: .custom)
-  open var currentAnnotations: [PeliasMapkitAnnotation : TGMapMarkerId] = Dictionary()
+  open var annotationToMarkerId: [PeliasMapkitAnnotation : TGMapMarkerId] = Dictionary()
```
msmollin: If I'm understanding the intent here, basically you want a reverse lookup in the dictionary (kind me the key associated for this value), correct? 

If that's the case, there's at least two different ways we can solve this with a single dictionary. We can write an extension on top of `Dictionary` to add this functionality, OR we can cast to `NSDictionary` which has more functionality.

The problem with both approaches is we can't assume this dictionary is bijective, so we have to return an array of possible results.

The NSDictionary approach looks like
`let keys = (dictionary as NSDictionary).allKeys(for: "myAwesomeValue") as! [String]` 
which you could then return the keys array. 

The pure swift approach looks something like this:
```
extension Dictionary where Value: Equatable {
  func keysForValue(value: Value) -> [Key] {
    return flatMap { (key: Key, val: Value) -> Key? in
      value == val ? key : nil
    }
  } 
}
```

I'm leaning towards pure swift because pure swift, but either would be good and more memory efficient than having two possibly large dictionaries hanging around.

karimnaaji: Agreed on the initial problem mentioned; though the pure swift solution is more memory efficient but lookups in one way would get expensive.

What about a `MapViewMarker` holding both marker ids? Those seems to always be coupled if I understand this correctly, it would then be possible to have single storage in a dictionary.

msmollin: Yeah the relationship is 1-1 so that could work.

FWIW `flatMap` is not THAT terrible for lookup performance. You can see the implementation of `flatMap` here - https://github.com/apple/swift/blob/master/stdlib/public/core/SequenceAlgorithms.swift.gyb#L673

I find it ironic that flatMap is literally just a for-in under the covers.

sarahlensing: I like the idea of an object to encapsulate both items but I will still have the original problem of only having either the marker id or the pelias annotation. Going to go with the dictionary extension for now and will keep an eye on the new TG marker work for an opportunity to refactor.

src/MapViewController.swift
sarahlensing: Nitpicks welcomed...this will be the last time I underuse the power of `guard`

src/MapViewController.swift
msmollin: Reviewed the dictionary extension usage - looks good. 2 Things:

1) Add unit tests for it to make sure it keeps behaving (especially in future swift versions. 4 is on the horizon!) .
2) `annotationToMarkerId` no longer makes a whole lot of sense I think. Any issue with putting it back to its original name?

sarahlensing: Added tests and reverted the dictionary name back to `currentAnnotations`

PR #181, opened 2017-02-24
Add map attribution

### LINKED ISSUE #87: Attribution

sarahlensing:
- Adds attribution button to bottom left of map which links to `https://mapzen.com/rights/` when clicked
- Fixes map's view height to consider tab bar
- Considers the new `mapView` property to be the main view and adds all other subview to it instead of to the view controller's view
- Fixes find me button positioning
- Adds string resource file
- Adds dimension resource file

img_4447

Closes #87


src/MapViewController.swift
```diff
@@ -68,7 +68,9 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
 
   //Error Domains for NSError Appeasement
   open static let MapzenGeneralErrorDomain = "MapzenGeneralErrorDomain"
+  private static let mapzenRights = "https://mapzen.com/rights/"
 
+  var application : ApplicationProtocol = UIApplication.shared
```
msmollin: Probably could be a `let` constant.

src/MapViewController.swift
```diff
@@ -446,6 +451,9 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
     findMeButton.setBackgroundImage(UIImage(named: "ic_find_me_pressed"), for: [.selected])
     findMeButton.backgroundColor = UIColor.white
     findMeButton.autoresizingMask = [.flexibleTopMargin, .flexibleLeftMargin]
+    let viewRect = mapView.bounds
```
msmollin: This won't work correctly. The `createFindMeButton()` function is called in the initializer of the VC. Referencing the view at this stage in the lifecycle will force view generation to happen outside of the proper lifecycle events. 

What are you attempting to solve by moving this around?

src/MapViewController.swift
```diff
@@ -487,6 +495,34 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
     tgViewController.markerRemove(marker)
     return
   }
+
+  //MARK: - private
+
+  private func setupTgControllerView() {
+    tgViewController.view.frame = CGRect(x: 0, y: 0, width: self.view.bounds.width, height: self.view.bounds.height-tabBarHeight)
```
msmollin: Couple issues:
- The reason you're needing to subtract the height here is because you're not using constraints. The view should adopt the size of its superview. So what I would do here is try here is to use the Autresizing Mask to keep it simple. Something like `.autoresizingMask = [.flexibleTopMargin, .flexibleLeftMargin, .flexibleRightMargin, .flexibleBottomMargin]` should do what you want. 

- The next issue is `self.view` doesn't always have a `bounds.width` in viewDidLoad, which is when this is being called. This is because of constraints. You really only get the right size of that once `viewDidLayoutSubviews` is finished. So, if you swap to using constraints, you shouldn't have to set the frame.

src/MapViewController.swift
```diff
@@ -487,6 +495,34 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
...
+  }
+
+  private func setupAttribution() {
+    attributionBtn = UIButton()
+    attributionBtn.setTitle(NSLocalizedString("attribution", comment: "Mapzen Attribution"), for: .normal)
+    attributionBtn.setTitleColor(.darkGray, for: .normal)
+    attributionBtn.titleLabel?.font = UIFont.systemFont(ofSize: 14)
+    attributionBtn.addTarget(self, action: #selector(openMapzenTerms), for: .touchUpInside)
+    attributionBtn.sizeToFit()
+    attributionBtn.translatesAutoresizingMaskIntoConstraints = false
```
msmollin: This should automatically get set to false when you add new constraints.

Localizable.strings
```diff
@@ -0,0 +1,9 @@
+/* 
+  Localizable.strings
```
msmollin: Did you check to make sure this gets imported when pulled in via cocoapods?

Dimens.swift
```diff
@@ -0,0 +1,14 @@
...
+//  Dimens.swift
+//  ios-sdk
+//
+//  Created by Sarah Lensing on 2/24/17.
+//  Copyright © 2017 Mapzen. All rights reserved.
+//
+
+import UIKit
+
+open class Dimens {
```
sarahlensing: 1. Public would make more sense...they probably wont subclass it, just access the constants
2. Updated
3. Renamed to Dimensions
4. Documented

src/MapViewController.swift
sarahlensing: Updated

src/MapViewController.swift
sarahlensing: I was focused on removing it from `viewWillAppear` since it doesnt need to be added every time the view appears but youre right, this isn't the place to add it. I've moved calling of this method from the initializer into `viewDidLoad`

src/MapViewController.swift
sarahlensing: The view's height doesn't adjust for the tab bar so I do need to adjust for it. I've updated setting of the `tgViewController` to use constraints.

src/MapViewController.swift
sarahlensing: Thats when constraints are set in IB:
```
If you want to use Auto Layout to dynamically calculate the size and position of your view, you must set this property to false, and then provide a non ambiguous, nonconflicting set of constraints for the view.
By default, the property is set to true for any view you programmatically create. If you add views in Interface Builder, the system automatically sets this property to false.
```

Localizable.strings
sarahlensing: No I didnt, will do that now

src/MapViewController.swift
msmollin: Got it. 👍

Localizable.strings
sarahlensing: Remove the strings file and will properly handle this in a follow up PR. Ticket here: https://github.com/mapzen/ios/issues/187

ios-sdkTests/MapViewControllerTests.swift
```diff
@@ -224,6 +226,7 @@ class MapViewControllerTests: XCTestCase {
   }
   
   func testFindMeButtonInitialState() {
+    controller.viewDidLoad()
```
msmollin: Mmmmm we should open the scope of `setupFindMeButton()` to `Internal` so we can test it. That'll still accomplish keeping it private from implementing devs.

ios-sdk.xcodeproj/project.pbxproj
```diff
@@ -142,29 +140,12 @@
 /* End PBXFrameworksBuildPhase section */
 
 /* Begin PBXGroup section */
-		7D7DAE171E60AFEE00FFCA6F /* resources */ = {
+		7D7847231E6775EA000D56CA /* dimensions */ = {
 			isa = PBXGroup;
 			children = (
-				7D7DAE191E60B00400FFCA6F /* dimens */,
-				7D7DAE181E60AFFE00FFCA6F /* strings */,
+				7D7847241E6775EA000D56CA /* Dimensions.swift */,
```
msmollin: Did this get moved into `src` on disk?

ios-sdkTests/MapViewControllerTests.swift
sarahlensing: Updated

ios-sdk.xcodeproj/project.pbxproj
sarahlensing: Yes. Its now in `src/dimensions/Dimensions.swift`

I'm thinking we might want to move our controllers, managers, extensions into separate folders in later PR too

PR #182, opened 2017-03-01
Update stylesheet

### LINKED ISSUE #174: Update BubbleWrap stylesheet version bundled with SDK

sarahlensing:
- Updates bundled stylesheet and related assets to `bubble-wrap-more-labels` `a6e0c3a`
- Removes method of queueing/applying scene updates to set api key
- Uses new scene update methods to configure global sdk api key variable when scene file is loaded

Closes #174


ios-sdkTests/MapViewControllerTests.swift
```diff
@@ -170,27 +170,35 @@ class MapViewControllerTests: XCTestCase {
...
     XCTAssertEqual(tgViewController.scenePath, "path")
-    XCTAssertEqual(tgViewController.sceneUpdates, updates)
+    XCTAssertEqual(tgViewController.sceneUpdates.count, 2)
+    XCTAssertTrue(tgViewController.sceneUpdates.contains(update))
+    XCTAssertEqual(tgViewController.sceneUpdates.last?.path, "global.sdk_mapzen_api_key")
   }
   
   func testLoadSceneFileAsync() {
-    controller.loadSceneFileAsync("path" , onSceneLoaded: nil)
+    try? controller.loadSceneFileAsync("path" , onSceneLoaded: nil)
```
msmollin: I guess this is an OK test? Really I think a more interesting test could be to use Xcode's asynchronous testing framework to guarantee that the load happens, but that would require a view hierarchy in place so probably we should eventually work towards some integration testing for this.

PR #198, opened 2017-03-07
Setup parent/child relationship with wrapped TGViewController

### LINKED ISSUE #184: MapViewController should make calls to TGMapViewController about becoming its parent view controller
Realized when looking over the code we're not properly becoming the parent of the TGMapViewController which will cause some issues depending on how devs wind up displaying MapViewController itself.

See implementation needed here:
https://developer.apple.com/library/content/featuredarticles/ViewControllerPGforiPhoneOS/ImplementingaContainerViewController.html#//apple_ref/doc/uid/TP40007457-CH11-SW13

sarahlensing:
- Creates parent/child relationship when wrapping `TGViewController`
- Adds test to ensure relationship exists

Closes #184

### DIFF
--- a/ios-sdkTests/MapViewControllerTests.swift
+++ b/ios-sdkTests/MapViewControllerTests.swift
@@ -41,6 +41,8 @@ class MapViewControllerTests: XCTestCase {
     controller.tgViewController = tgViewController
     let mockHTTP = MockHTTPHandler()
     controller.tgViewController.httpHandler = mockHTTP
+
+    let _ = controller.view
   }
 
   func testInit() {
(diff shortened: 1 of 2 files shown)


ios-sdkTests/MapViewControllerTests.swift
```diff
@@ -41,12 +41,18 @@ class MapViewControllerTests: XCTestCase {
     controller.tgViewController = tgViewController
     let mockHTTP = MockHTTPHandler()
     controller.tgViewController.httpHandler = mockHTTP
+
+    let _ = controller.view
```
msmollin: Swift 3 changed this - you can just do `_ = controller.view`

PR #200, opened 2017-03-08
Pass api key via commandline

### LINKED ISSUE #195: Circle CI nightly builds no longer display a map
The combination of removing the hard-coded Mapzen API key from the demo app (good) and turning off keyless access to vector tiles (good) means that nightly builds on Circle CI now fail to load a map (bad).

We need to come up with a way to securely inject a demo API key into the iOS sample app at build time on Circle CI. Perhaps some combination of a Circle CI environment variable and build properties.

sarahlensing:
- Passes api key to `xcodebuild`
- Updates key to be read from `Info.plist`
- Adds `MAPZEN_API_KEY` to circle environment variables
- Adds ability to create custom, non-shared scheme and pass api key environment variable for local development

To test this locally I added a test to `MapViewControllerTests`:
```
  func testKey() {
    let apiKey = Bundle.main.infoDictionary?["MAPZEN_API_KEY"] as? String
    XCTAssertEqual(apiKey, "theApiKey")
  }
```

and ran: 
```
xcodebuild
        -sdk iphonesimulator
        -workspace ios-sdk.xcworkspace
        -scheme "ios-sdk"
        -destination 'platform=iOS Simulator,name=iPhone 6,OS=latest'
        MAPZEN_API_KEY=$MAPZEN_API_KEY
        clean build test
```

I also duplicated the `ios-sdk` scheme:
screen shot 2017-03-08 at 10 18 00 am

and added the api key as an environment variable (since I don't run xcodebuild when doing local dev):
screen shot 2017-03-08 at 10 17 42 am

Closes #195

### DIFF
--- a/SampleApp/AppDelegate.swift
+++ b/SampleApp/AppDelegate.swift
@@ -16,8 +16,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
   func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
     // Override point for customization after application launch.
-    let apiKey = Bundle.main.infoDictionary?["MAPZEN_API_KEY"] as? String
-    MapzenManager.sharedManager.apiKey = apiKey
+    let mapzenKeyName = "MAPZEN_API_KEY"
+    guard let apiKey = Bundle.main.infoDictionary?[mapzenKeyName] as? String else { return true }
+    if apiKey.isEmpty {
+      MapzenManager.sharedManager.apiKey = ProcessInfo.processInfo.environment[mapzenKeyName]
+    } else {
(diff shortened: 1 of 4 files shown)


ecgreb:
Cool it looks like this fixes our issue on Circle CI and for the nightly builds.

How does one set the Mapzen API key if the demo app is installed via `pod try Mapzen-ios-sdk`?

sarahlensing:
They would have two ways to do this:
1. Set this value directly in the SampleApp/Info.plist (replacing the variable that is there)
2. Create a custom scheme and a MAPZEN_API_KEY environment variable

The first way is quicker but will create a diff that they will have to be sure not to check in. So, if they are just downloading to test things out and not commit code this would be the way to go. The second way requires a few more steps but the changes will be transparent to git...a more longterm solutions if they are going to be doing ongoing development.

PR #211, opened 2017-03-13
Show status bar in landscape orientation

sarahlensing:
- Updates sample app controllers to always show the status bar and reinforce standard iOS UI (inspired by one-off request from John for blog post screenshot)

### DIFF
--- a/SampleApp/AutocompleteTableVC.swift
+++ b/SampleApp/AutocompleteTableVC.swift
@@ -6,12 +6,11 @@
 //  Copyright © 2016 Mapzen. All rights reserved.
 //
 
-import UIKit
 import MapKit
 import CoreLocation
 import Pelias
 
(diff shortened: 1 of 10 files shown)


msmollin:
Hmmmm this is technically non-standard iOS UI these days for phones. In landscape, Apple generally recommends hiding the status bar because it shows more content. On iPad, the status bar is still generally shown because of the additional space available on tablets. 

See example in: https://developer.apple.com/ios/human-interface-guidelines/visual-design/layout/

msmollin:
That said if its something we really want, I'm OK with this.

sarahlensing:
I think that for now, because we want this for screenshots and because the app's content doesn't feel squished, we want this. We can revisit it if the app becomes too crowded.

msmollin:
Very good. Agreed.

PR #228, opened 2017-03-20
Update circle.yml

kkowalsky:
"tickler" is now named beacon


msmollin:
@kkowalsky this won't build right until we fix the issue over here: https://github.com/mapzen/ios/pull/230

kkowalsky:
no worries @msmollin!

msmollin:
We're all good in master. Feel free to rebase this and it should build. You may need to do a rebuild without cache to kick it.

PR #236, opened 2017-03-21
Adds low memory management to MapViewController and fixes some other bugs

### LINKED ISSUE #222: Switching tabs in demo app breaks map?
Memory issue?
Find me button tends to quit first.
Switching to routing tab makes basic map demo tab turn black (sometimes).
`didReceiveMemoryWarning` should be better handled in Tangram

### LINKED ISSUE #235: Computed properties for tilt/float/etc. that pass values into TGViewController are broken
Current code:
```  
open var rotation: Float {
    set {
      tgViewController.rotation = rotation
    }
    get {
      return tgViewController.rotation
    }
  }
```

`= rotation` on this side is actually pulling from the TGViewController's current `rotation` value, so its effectively always zero and ignores the value.

It really should be:

```tgViewController.rotation = newValue```

### PR DESCRIPTION
**Pull request by msmollin**
### Overview
MapViewController now tries to recover from a low memory warning.
There were also some issues with properties not behaving as expected
that were fixed as part of this.

### Proposed Changes
Implement a management layer for low memory situations so we can save state and then attempt to rebuild the state when coming back on screen. Tested on an iPhone 6S+ it seems to behave well, but could use more testing on older / slower devices.

Fixes #235 
Fixes #222

### DIFF
--- a/src/MapViewController.swift
+++ b/src/MapViewController.swift
@@ -822,18 +822,7 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
     shouldFollowCurrentLocation = button.isSelected
   }
 
-  // MARK:- ViewController Lifecycle
-
-  override open func viewDidLoad() {
-    super.viewDidLoad()
-    locationManager.delegate = self
-    setupTgControllerView()
-    setupAttribution()
-    setupFindMeButton()
-
-    tgViewController.gestureDelegate = self
-    tgViewController.mapViewDelegate = self
-  }
+  // MARK:- Memory Management handlers
 
   func reloadTGViewController() {
     guard let unwrappedSaver = stateSaver else { return }
@@ -851,7 +840,9 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
     }
   }
 
+  //This function should really only be used in conjunction with reloadTGViewController, but it is safe to be used whenever.
   func recreateMap(_ state: StateReclaimer) {
+    //Annotation Replay
     let oldAnnotations = self.currentAnnotations
     self.currentAnnotations = Dictionary()
     for (annotation, marker) in oldAnnotations {
@@ -860,6 +851,8 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
       newMarker.stylingString = marker.stylingString
       self.currentAnnotations[annotation] = newMarker
     }
+
+    //Routing Replay
     if let currentRoute = self.currentRoute {
       do {
         try self.display(currentRoute)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/MapViewController.swift [hunk 1]
```diff
@@ -813,6 +834,81 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
...
+        //Silently catch this - we may still be able to recover from here sans route
+      }
+    }
+    self.tilt = state.tilt
+    self.rotation = state.rotation
+    self.zoom = state.zoom
+    self.position = state.position
+    self.cameraType = state.cameraType
+    self.setupAttribution()
+    self.setupFindMeButton()
```
sarahlensing: We also need to check to see if we should show the find me button and add:
```swift
if shouldShowCurrentLocation {
  _ = self.showCurrentLocation(true)
}
```

Maybe we also want to add a similar way to reshow the current location marker?

src/MapViewController.swift
```diff
@@ -813,6 +834,81 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
...
+  }
+
+  override open func viewWillAppear(_ animated: Bool) {
+    super.viewWillAppear(animated)
+    //We only want to attempt to reload incase we reloaded due to memory issues
+    if (!tgViewController.isViewLoaded) {
```
msmollin: I went back and forth on that. My only concern there was if we wind up using state saver for other things but I do agree it would be more resilient not to rely on Tangram. Will update.

src/MapViewController.swift
msmollin: Yep. Oversight on my part on those two. Will update.

src/MapViewController.swift
```diff
@@ -813,6 +893,40 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
...
+  override open func didReceiveMemoryWarning() {
+    if (!isCurrentlyVisible) {
+      tgViewController.willMove(toParentViewController: nil)
+      tgViewController.view.removeConstraints(tgViewController.view.constraints)
+      tgViewController.view.removeFromSuperview()
+      tgViewController.removeFromParentViewController()
+      stateSaver = StateReclaimer(tilt: tilt, rotation: rotation, zoom: zoom, position: position, cameraType: cameraType, mapStyle: currentStyle)
+      //Probably unnecessary, but just incase we don't want any calls coming through from older versions that the OS has yet to clean up
+      tgViewController.gestureDelegate = nil
+      tgViewController.mapViewDelegate = nil
```
sarahlensing: The current location gem is still not showing for me. We need to remove it from the current `tgViewController` before creating a new one since they can only be attached to one map at a time (according to the tangram docs)

We should add:
```
currentLocationGem?.map = nil
```
after nil'ing out hte delegates here and then add `currentLocationGem?.map = tgViewController` when we restore here:
```
if shouldShowCurrentLocation {
  currentLocationGem?.map = tgViewController
  _ = self.showCurrentLocation(true)
}
```

src/MapViewController.swift
```diff
@@ -801,10 +821,70 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
...
+
+    // Location Marker reset
+    if shouldShowCurrentLocation {
+      _ = self.showCurrentLocation(true)
+    }
+
+    //Button Setup
+    self.setupAttribution()
+    let originalButton = findMeButton
+    self.setupFindMeButton()
```
sarahlensing: We also need to re-enable the button if the original button was enabled

msmollin: Bah yep. Forgot to test that. Updated.

src/MapViewController.swift
msmollin: Think I have this working now. I think this broke as part of the refactor to TGMarker but I could be wrong.

msmollin:
@sarahlensing included a rebase in this, but I think I got your last two comments in 21b1997 - let me know!

Issue #237, opened 2017-03-21
Document MapzenSearch API


sarahlensing:
- Document APIs and add samples to README & website

msmollin:
Completed: https://mapzen.com/documentation/ios/search/

PR #238, opened 2017-03-22
Wrap Search Configs

### LINKED ISSUE #80: Wrap pelias search config structs
Enables searching from objective-c

sarahlensing:
### Overview
This adds objective-c support for part of the Pelias API. A follow-up PR will integrate these changes into a new `MapzenSearch` object and complete objective-c compatibility.

### Proposed Changes
- Wrap `Pelias` structs in `NSObject` equivalents
- Wrap and unwrap these objects using `MapzenSearchDataConverter`

Future work:
- Documentation https://github.com/mapzen/ios/issues/237
- Expose properties on wrapper objects (dependent on): https://github.com/pelias/pelias-ios-sdk/issues/57
- Remove partial objective-c compatibility work in Pelias: https://github.com/pelias/pelias-ios-sdk/issues/56

Closes #80


src/MapzenResponse.swift
```diff
@@ -0,0 +1,52 @@
...
+//  ios-sdk
+//
+//  Created by Sarah Lensing on 3/21/17.
+//  Copyright © 2017 Mapzen. All rights reserved.
+//
+
+import Foundation
+import Pelias
+
+public class MapzenResponse : NSObject {
```
msmollin: This should probably be something more descriptive like `SearchResponse`. 

Also it's not practice in swift to prefix names with things like `MZ` or `Mapzen` because namespaces are implicit. So if there's a namespace conflict you would use `ModuleName.Class.Function()` instead of `Class.Function()`. So in general I would remove that prefix from these classes / structs / etc. This is something I failed to understand while building Pelias, so lets fix that here :) 

There are always exceptions however. For example the `MapzenManager` is a good one since it's a state manager for Mapzen services, but otherwise I think we can be a little more loose with naming.

src/MapzenPlaceConfig.swift
```diff
@@ -0,0 +1,28 @@
...
+//  Pods
+//
+//  Created by Sarah Lensing on 3/20/17.
+//
+//
+
+import Foundation
+import Pelias
+
+public class MapzenPlaceConfig : NSObject {
```
msmollin: What are you thoughts on moving all of these into a single swift file? It's common practice to have common structs / class grouped in the same file in Swift.

src/MapzenSearchDataConverter.swift
```diff
@@ -0,0 +1,153 @@
+//
+//  MapzenSearchDataConverter.swift
+//  Pods
```
msmollin: Nit: Just noticed the `Pods` name here. This should be the project name, so `Mapzen-ios-sdk` or something similar.

src/MapzenSearchDataObjects.swift
```diff
@@ -0,0 +1,103 @@
...
+//  Pods
+//
+//  Created by Sarah Lensing on 3/20/17.
+//
+//
+
+import Foundation
+import Pelias
+
+public class MzSearchRect: NSObject {
```
msmollin: Yeah like I said above we can probably remove all of the prefixes here and then for things like `SearchBoundaryCircle` which will conflict on the namespace you should be able to reference it by module name like `Pelias.SearchBoundaryCircle`

ios-sdkTests/MapzenSearchDataObjectsTests.swift
```diff
@@ -0,0 +1,47 @@
...
+    let rect = SearchBoundaryRect(minLatLong: MapzenSearchDataConverter.unwrapPoint(min), maxLatLong: MapzenSearchDataConverter.unwrapPoint(max))
+    XCTAssertEqual(searchRect.rect, rect)
+    XCTAssertEqual(searchRect, MzSearchRect(minLatLong: min, maxLatLong: max))
+  }
+
+  func testSearchCircle() {
+    let center = MzGeoPoint(latitude: 70.0, longitude: 40.0)
+    let searchCircle = MzSearchCircle(center: center, radius: 8)
+    let circle = SearchBoundaryCircle(center: MapzenSearchDataConverter.unwrapPoint(center), radius: 8)
+    XCTAssertEqual(searchCircle.circle, circle)
```
msmollin: You should probably use `XCTAssertEqualWithAccuracy()` here and for the `accuracy` part use `DBL_EPSILON` to avoid possible compiler weirdness with floats, since checking equality with floats tends to be hit-or-miss.

ios-sdkTests/MapzenSearchDataConverterTests.swift
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  func testWrapQueryItem() {
+    let item = PeliasPlaceQueryItem.init(placeId: "id", dataSource: .OpenStreetMap, layer: .venue)
+    let wrapped = MapzenSearchDataConverter.wrapQueryItem(item)
+    XCTAssertEqual(wrapped.placeId, "id")
+    XCTAssertEqual(wrapped.dataSource, .openStreetMap)
+    XCTAssertEqual(wrapped.layer, .venue)
+  }
+
```
msmollin: Nit: whitespace.

src/MapzenPlaceConfig.swift
sarahlensing: Its against what inherently feels natural to me (its overwhelming to see large files) but if its swift common practice I'll move it into one file.

src/MapzenSearchDataConverter.swift
sarahlensing: Will update to `ios-sdk`

ios-sdkTests/MapzenSearchDataObjectsTests.swift
sarahlensing: I thought that weirdness was just when the float was returned from TGMapViewController's `getPosition`?

ios-sdkTests/MapzenSearchDataConverterTests.swift
sarahlensing: I'm not seeing whitespace on this line:
screen shot 2017-03-23 at 2 16 19 pm

Was it on a different line?

src/MapzenSearchDataObjects.swift
sarahlensing: How do you feel about changing the sdk module name from `ios_sdk` to `Mapzen`?

ios-sdkTests/MapzenSearchDataObjectsTests.swift
msmollin: Indeed, but in general you shouldn't do direct equality checks on floats without an accuracy check in place (which is why I started down that path in the first place).

PR #243, opened 2017-03-27
Add Marker Class

### LINKED ISSUE #221: Create SDK Marker Class
### Description
Create SDK Marker class to wrap TGMarker

- Add convenience methods: `isVisible`, `setSize`, `setBackgroundColor`, `setInteractive` (having it default to true) and then handle concatenating values to internal style string
- Provide system marker types
    - Like UIBarButtonItem's - (instancetype)initWithBarButtonSystemItem:(UIBarButtonSystemItem)systemItem target:(nullable id)target action:(nullable SEL)action; we can provide an easy way for developers to add markers with sdk styles such as sdk-point-overlay

sarahlensing:
### Overview
Adds `Marker` class which allows developers to set basic properties without having to manipulate style strings. If they would like to do advanced work with style strings or paths, the underlying `TGMarker` class is available

### Proposed Changes
- Adds `Marker` class with properties `size`, `backgroundColor`, and `interactive` to handle updating `stylingString`
- Adds `MarkerType` for common layers: `currentLocation`, `searchPin`, and `routeLine`
- Updates `MapController` to use `Marker`

Closes #221

### DIFF
--- a/src/MapViewController.swift
+++ b/src/MapViewController.swift
@@ -481,6 +481,24 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
     tgViewController.animate(toTilt: radians, withDuration: seconds, with: easeType)
   }
 
+  /**
+   Adds a marker to the map. A marker can only be added to one map at a time.
+   
+   - parameter marker: The marker to add to map.
+   */
+  open func addMarker(_ marker: Marker) {
+    marker.tgMarker.map = tgViewController
+  }
+
+  /**
+   Removes a marker from the map.
+
+   - parameter marker: The marker to remove from map.
+   */
+  open func removeMarker(_ marker: Marker) {
+    marker.tgMarker.map = nil
+  }
+
   /// Removes all existing markers on the map.
   open func markerRemoveAll() {
     tgViewController.markerRemoveAll()
--- a/src/MapViewController.swift
+++ b/src/MapViewController.swift
@@ -871,8 +871,11 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
       addMarker(newMarker)
       //TODO: also set polyline, polygon etc
       newMarker.point = marker.point
-      newMarker.tgMarker.stylingString = marker.stylingString
-      newMarker.tgMarker.stylingPath = marker.stylingPath
+      if !marker.stylingPath.isEmpty {
+        newMarker.tgMarker.stylingPath = marker.stylingPath
+      } else {
+        newMarker.tgMarker.stylingString = marker.stylingString
+      }
       self.currentAnnotations[annotation] = newMarker.tgMarker
     }
 
--- a/src/MapViewController.swift
+++ b/src/MapViewController.swift
@@ -180,10 +180,10 @@ public protocol MapMarkerSelectDelegate : class {
    Informs the delegate a marker of the map was just selected
 
    - parameter controller: The MapViewController that recognized the selection.
-   - parameter markerPickResult: A marker selection returned as an instance of TGMarkerPickResult.
+   - parameter markerPickResult: A marker selection returned as an instance of Marker.
    - parameter atScreenPosition: The screen coordinates of the picked marker.
    */
-  func mapController(_ controller: MapViewController, didSelectMarker markerPickResult: TGMarkerPickResult, atScreenPosition position: CGPoint)
+  func mapController(_ controller: MapViewController, didSelectMarker marker: Marker, atScreenPosition position: CGPoint)
 }
 
 /// MapTileLoadDelegate
@@ -223,6 +223,7 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
   open var shouldFollowCurrentLocation = false
   open var findMeButton = UIButton(type: .custom)
   var currentAnnotations: [PeliasMapkitAnnotation : TGMarker] = Dictionary()
+  var currentMarkers: [TGMarker : Marker] = Dictionary()
   open var attributionBtn = UIButton()
   private var locale = Locale.current
   var stateSaver: StateReclaimer?
@@ -487,6 +488,7 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
    - parameter marker: The marker to add to map.
    */
   open func addMarker(_ marker: Marker) {
+    currentMarkers[marker.tgMarker] = marker
     marker.tgMarker.map = tgViewController
   }
 
@@ -496,6 +498,7 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
    - parameter marker: The marker to remove from map.
    */
   open func removeMarker(_ marker: Marker) {
+    currentMarkers.removeValue(forKey: marker.tgMarker)
     marker.tgMarker.map = nil
   }
 
@@ -1106,9 +1109,11 @@ extension MapViewController : TGMapViewDelegate, TGRecognizerDelegate {
   
   open func mapView(_ mapView: TGMapViewController, didSelectMarker markerPickResult: TGMarkerPickResult?, atScreenPosition position: CGPoint) {
     guard let markerPickResult = markerPickResult else { return }
-    let marker = markerPickResult.marker
-    guard let annotation = currentAnnotations.keyForValue(value: marker) else {
-      markerSelectDelegate?.mapController(self, didSelectMarker: markerPickResult, atScreenPosition: position)
+    let tgMarker = markerPickResult.marker
+    guard let annotation = currentAnnotations.keyForValue(value: tgMarker) else {
+      if let marker = currentMarkers[tgMarker] {
+        markerSelectDelegate?.mapController(self, didSelectMarker: marker, atScreenPosition: position)
+      }
       return
     }
     if let target = annotation.target, let action = annotation.selector {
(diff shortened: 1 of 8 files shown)


src/MapViewController.swift
```diff
@@ -210,19 +210,20 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
...
+  var currentLocationGem: Marker?
   var lastSetPoint: TGGeoPoint?
   var shouldShowCurrentLocation = false
-  var currentRouteMarker: TGMarker?
+  var currentRouteMarker: Marker?
   var currentRoute: OTRRoutingResult?
   open var shouldFollowCurrentLocation = false
   open var findMeButton = UIButton(type: .custom)
   var currentAnnotations: [PeliasMapkitAnnotation : TGMarker] = Dictionary()
+  var currentMarkers: [TGMarker : Marker] = Dictionary()
```
karimnaaji: Do you need that to be exposed in `TGMapVIewController` instead?

src/MapViewController.swift
```diff
@@ -846,10 +867,13 @@ open class MapViewController: UIViewController, LocationManagerDelegate {
...
     for (annotation, marker) in oldAnnotations {
-      let newMarker = TGMarker.init(mapView: self.tgViewController)
+      let newMarker = Marker.init()
+      addMarker(newMarker)
+      //TODO: also set polyline, polygon etc
       newMarker.point = marker.point
-      newMarker.stylingString = marker.stylingString
-      self.currentAnnotations[annotation] = newMarker
+      newMarker.tgMarker.stylingString = marker.stylingString
+      newMarker.tgMarker.stylingPath = marker.stylingPath
```
karimnaaji: `stylingString` and `stylingPath` are mutually exclusive. If you give a marker a styling string this one would be used, if you give it a path to a draw rule, the style of this targeted draw rule will be used. 

Refer implementation details in https://github.com/tangrams/tangram-es/pull/1242 and the notes in https://github.com/tangrams/tangram-es/blob/master/platforms/ios/src/TangramMap/TGMarker.h#L67.

src/Marker.swift
```diff
@@ -0,0 +1,194 @@
...
+
+import Foundation
+import TangramMap
+
+public enum MarkerType : Int {
+  case currentLocation, searchPin, routeLine
+}
+
+@objc(MZMarker)
+public class Marker: NSObject {
```
msmollin: This is a good start. I would generalize this to define a Marker Protocol that this class would then implement. You could even move much of that implementation into a protocol extension if you wanted.

Then, change out the implementing call sites for this class to use the protocol object instead of the actual class itself.
Thoughts?

src/Marker.swift
```diff
@@ -0,0 +1,194 @@
...
+  // all marker types have an associated styling path
+  private static let typeToStylingPath = [MarkerType.currentLocation : "layers.mz_current_location_gem.draw.ux-location-gem-overlay",
+                                   MarkerType.searchPin : "layers.mz_search_result.draw.ux-icons-overlay",
+                                   MarkerType.routeLine : "layers.mz_route_line.draw.ux-route-line-overlay"]
+  //currently only search results have an inactive state
+  private static let typeToInactiveStylingPath = [MarkerType.searchPin : "layers.mz_search_result.inactive.draw.ux-icons-overlay"]
+
+  open var point: TGGeoPoint? {
+    set {
+      guard let p = newValue else { return }
```
msmollin: How do you remove the point later then if we return here?

src/Marker.swift
```diff
@@ -0,0 +1,194 @@
...
+
+@objc(MZMarker)
+public class Marker: NSObject {
+
+  private static let kPointStyle = "points"
+  private static let kLineStyle = "lines"
+  private static let kPolygonStyle = "polygons"
+  private static let kDefaultBackgroundColor = UIColor.white
+  private static let kDefaultInteractive = true
+  private static let kDefaultSize = CGSize.zero
```
karimnaaji: 0 seems a little bit aggressive as a default, if the user sets everything but that, the marker wouldn't show.

src/MapViewController.swift
karimnaaji: Agreed that this could easily go away, I think a good pattern here would be to be able to set a `userData` directly to a `TGMarker`:

```objc
- (void)userData:(void*)userData
```

so that the user could set whatever data it needs to it, and retrieve it in a callback whenever a selection happens, so storing objects mapping like this wouldn't be necessary.

tallytalwar: Should this `userData` on a marker be set in the core tangram interface in that case?

karimnaaji: Yep, I think this would make things easier here.

msmollin: That would align with parts of Apple's Mapkit. I don't think Google allows for storage of random data on a marker.

karimnaaji: It seems like it does: https://developers.google.com/maps/documentation/ios-sdk/reference/interface_g_m_s_overlay.html#a900ff2be989386d501ed6c64ea1d7ab3

src/MapViewController.swift
sarahlensing: Updated to check if the path is empty and if not set it, otherwise, set the styling string since `Marker` objects will have one or the other set (depending on the constructor used)

src/Marker.swift
sarahlensing: This is how most UI components work in iOS (`UIButton`, `UIImage`). Unless you use the constructor which takes a size or image, it is zero. I think its okay to keep this as is because developers are used to other components behaving this way.

src/Marker.swift
```diff
@@ -0,0 +1,194 @@
...
+      return tgMarker.point
+    }
+  }
+
+  open var polyline: TGGeoPolyline? {
+    set {
+      guard let l  = newValue else { return }
+      tgMarker.polyline = l
+      styleType = Marker.kLineStyle
+      updateStyleString()
```
sarahlensing: Nice catch, updated.

src/Marker.swift
sarahlensing: I had this in a local TODO...good question...

The underlying `TGMarker` property is non-optional. @karimnaaji @tallytalwar Should I make a ticket to allow setting nil values or is there a recommended way to handle this?

src/Marker.swift
sarahlensing: Sounds like a nice change, will update now

src/MapViewController.swift
sarahlensing: Created tickets for this: 
https://github.com/tangrams/tangram-es/issues/1409
https://github.com/mapzen/ios/issues/246

src/Marker.swift
sarahlensing: Oh wait...objective-c protocols can't have properties...

msmollin: Uhhhh yes they can?

https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithProtocols/WorkingwithProtocols.html

`Protocols can include declarations for both instance methods and class methods, as well as properties.`

src/Marker.swift
karimnaaji: The problem here is a pixel point with a style size defined to be 0 has a specific meaning in Tangram for point labels https://github.com/tangrams/tangram-es/issues/1193. 

Also, specifying a size at all has a meaning, if you have a point sprite (an icon) with no size but it is defined in the stylesheet like so: https://github.com/tangrams/tangram-es/blob/master/scenes/scene.yaml#L45, the size of this default point icon would be used by default (here `32px`x`32px`), unless you specify a `size` parameter in the style. 

So, thinking more of this, It seems to me that the best option is to only provide the size in the style string if it has been provided by the user without assuming a default?

src/Marker.swift
karimnaaji: TGGeoPoint is a C-struct, I think this is not necessary, is this code testing whether it is `nil`?

src/Marker.swift
msmollin: What happens in the case where a size isn't provided by the marker, and a size isn't provided in the stylesheet?

karimnaaji: @msmollin In your case it's not possible that the stylesheet doesn't provide a size, since you use the overlay sprites referenced in the stylesheet.

The case you need to handle though is the user providing a `UIImage` but not size, in that case I would user the size of the `UIImage` as a potential default.

src/Marker.swift
msmollin: Yeah because it's an optional.

I probably attached my comment to the wrong thing here - but the basic question I have is for all of these (like the poly lines and polygons) is if we want to remove the poly line what's the correct way to go about it? Setting `nil` seems to be the correct way in iOS land, but would we rather nil out the entire marker if that's the case?

karimnaaji: You can replace the marker geometry by any new geometry.

karimnaaji: I see the pattern here, we don't provide direct way to remove the geometry from it, only replacement. You can adjust the visibility of it to make it not visible but still usable in the current map, or remove the marker from the map view and then use it again later in the same or a different view.

src/Marker.swift
sarahlensing: Updated to include generic protocol

src/Marker.swift
sarahlensing: Updated to not pass the size if it isn't set by the user

msmollin:
After discussions with @sarahlensing we're gonna refactor this marker class to be broken up into smaller versions that more semantically describe their intended functions. Merging this for now so we can get the tests and such in place to make sure the refactor goes nice and smooth.

Issue #224, opened 2017-03-15
Sample app cleanup


msmollin:
Map tab:
- [x] Remove marker test button
- [x] Move the map language button into the map tab from search

Search Tab
- [x] Convert search bar to use autocomplete list instead of current /search endpoint
- [x] Selecting a search result places an annotation on the map 
- [x] Annotation placement also zooms in on that pin with a zoom level that makes sense
- [x] Annotation to be selectable and shows alert of name of place on selection
- [x] Remove map language button
- [x] Add find me button

Route Tab:
- [x] Use current location to start route
- [x] Zoom to current location upon start
- [x] Tap to route from current location to tap point on map
- [x] Remove search
- [x] Add find me button 
- [x] Move results button to left justified on screen

Issue #74, opened 2016-10-10
Swift sdk classes should follow common naming pattern


karimnaaji:
Refer https://github.com/mapzen/ios/pull/73 for related comments

PR #244, opened 2017-03-28
Refactor some class names and begin moving sample app functionality

### LINKED ISSUE #172: Rename TangramVC to TangramViewController
Per some outside feedback and previous feedback from @karimnaaji we should refactor TangramVC to be more explicitly declared.

msmollin:
### Overview

First step in #224 work as well as #74 

Fixes #172 by renaming to something more in line with the demo app

### DIFF
--- a/SampleApp/DemoMapViewController.swift
+++ b/SampleApp/DemoMapViewController.swift
@@ -8,20 +8,10 @@
 
 import UIKit
 import TangramMap
-class TangramVC:  SampleMapViewController, MapMarkerSelectDelegate {
+class DemoMapViewController:  SampleMapViewController, MapMarkerSelectDelegate {
 
   private var styleLoaded = false
   private var markerVisible = false
-
-  lazy var testMarker : TGMarker = { [unowned self] in
-    let marker = TGMarker.init(mapView: self.tgViewController)
-    marker.stylingString = "{ style: 'points', color: 'white', size: [50px, 50px], collide: false, interactive: true }"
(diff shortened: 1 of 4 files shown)


SampleApp/DemoMapViewController.swift
```diff
@@ -102,4 +85,34 @@ class TangramVC:  SampleMapViewController, MapMarkerSelectDelegate {
       self.activityIndicator.stopAnimating()
     })
   }
+
+  @objc private func changeMapLanguage() {
+    let actionSheet = UIAlertController.init(title: "Map Language", message: "Choose a language", preferredStyle: .actionSheet)
+    actionSheet.addAction(UIAlertAction.init(title: "English", style: .default, handler: { [unowned self] (action) in
```
sarahlensing: Not your code but we could update the `addAction` calls here to be:

```
let languageIdByActionSheetTitle = [
      "English": "en-US",
        ...
      ]
let actionSheet = UIAlertController.init(title: "Map Language", message: "Choose a language", preferredStyle: .actionSheet)
for (actionTitle, languageIdentifier) in languageIdByActionSheetTitle {
    actionSheet.addAction(UIAlertAction.init(title: actionTitle, style: .default, handler: { [unowned self] (action) in
      self.updateLocale(Locale.init(identifier: languageIdentifier))
  }))
}
```

msmollin: Cool I'll do that in the next PR.

PR #245, opened 2017-03-28
Add MapzenSearch

### LINKED ISSUE #79: Wrap PeliasSearchManager to enable obj-c interoperability
PeliasSearchManager doesn't show up in Objective-C because it uses Swift structs, and as such the compiler won't import it.

sarahlensing:
### Overview
Completes objective-c compatibility work by wrapping `PeliasSearchManager` and related operation objects in `MapzenSearch`

### Proposed Changes
- Create main entry point for search via `MapzenSearch`
- Fix pod source files
- Update `Pelias` dependency (https://github.com/pelias/pelias-ios-sdk/pull/58)

Closes #79

### DIFF
--- a/ios-sdkTests/MapzenSearchTests.swift
+++ b/ios-sdkTests/MapzenSearchTests.swift
@@ -1,0 +1,276 @@
+//
+//  MapzenSearchTests.swift
+//  ios-sdk
+//
+//  Created by Sarah Lensing on 3/28/17.
+//  Copyright © 2017 Mapzen. All rights reserved.
+//
+
+import XCTest
+@testable import ios_sdk
+import Pelias
+
+public class MapzenSearchTests : XCTestCase {
+
+  let mapzenSearch = MapzenSearch.sharedInstance
+  let peliasSearchManager = PeliasSearchManager.sharedInstance
+
+  func testAutocompleteTimeDelay() {
+    mapzenSearch.autocompleteTimeDelay = 3
+    XCTAssertEqual(3, mapzenSearch.autocompleteTimeDelay)
+    XCTAssertEqual(3, peliasSearchManager.autocompleteTimeDelay)
+  }
+
+  func testBaseUrl() {
+    let defaultBaseUrl = URL.init(string: Constants.URL.base)
+    XCTAssertEqual(mapzenSearch.baseUrl, defaultBaseUrl)
+    XCTAssertEqual(peliasSearchManager.baseUrl, defaultBaseUrl)
+
+    let baseUrl = URL.init(string: "https://search.custom.com")!
+    mapzenSearch.baseUrl = baseUrl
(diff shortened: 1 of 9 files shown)


src/MapzenSearch.swift
```diff
@@ -0,0 +1,216 @@
...
+      peliasOperation.name = newValue
+    }
+  }
+
+}
+
+@objc(MZSearchError)
+public class SearchError: NSObject {
+
+  let peliasError: PeliasError
```
msmollin: Hmmm we already have an MZError class over in map view controller. I wonder if there's a way to use that here instead of wrapping the pelias errors?

ios-sdkTests/MapzenSearchTests.swift
```diff
@@ -0,0 +1,276 @@
...
+//  Copyright © 2017 Mapzen. All rights reserved.
+//
+
+import XCTest
+@testable import ios_sdk
+import Pelias
+
+public class MapzenSearchTests : XCTestCase {
+
+  let mapzenSearch = MapzenSearch.sharedInstance
```
sarahlensing: Didn't know that...Updated.

ios-sdkTests/MapzenSearchTests.swift
```diff
@@ -0,0 +1,276 @@
...
+    let config = PlaceConfig.init(places: [PlaceQueryItem.init(placeId: "", dataSource: .openAddresses, layer: .address)]) { (response) in
+      //
+    }
+    let operation = mapzenSearch.placeQuery(config)
+    XCTAssertNotNil(operation)
+  }
+}
+
+class SearchOperationTests : XCTestCase {
+  let operation = SearchOperation.init(TestPeliasOperation())
```
sarahlensing: Updated.

ios-sdkTests/MapzenSearchTests.swift
```diff
@@ -0,0 +1,276 @@
...
+
+  func testStart() {
```
sarahlensing: Updated to call `main()` but FYI the `PeliasSearchOperation` is a stub and does no real work in either `start()` or `main()`

src/MapzenSearch.swift
```diff
@@ -0,0 +1,216 @@
...
+@objc(MZSearchOperation)
+public class SearchOperation : Operation {
+
+  let peliasOperation: PeliasOperation
+
+  init(_ op: PeliasOperation) {
+    peliasOperation = op
+  }
+
+  override public func start() {
```
sarahlensing: I wrapped it because I didnt want to recreate a new operation given the `PeliasSearchOperation` that is returned by the `PeliasSearchManager`. It seemed more complicated (I'd have to add it to the internal queue and remove the returned operation from that queue).

Issue #249, opened 2017-03-31
Support more system marker types


sarahlensing:
We support a subset of the available system markers defined in the house stylesheets.

Add support for these types:
transit stops, shields, transit lines, route lines, default lines, and default polygons

Potentially consider dashed lines (test how they look though...last time we used them they were jagged)

PR #250, opened 2017-03-31
Split marker protocol

sarahlensing:
### Overview
This work splits the `GenericMarker` protocol into several smaller, more specific protocols. It also creates base implementations for each protocol. In addition, it adds style properties which prevented polygons and lines from showing on the map.

### Proposed Changes
- There are now 6 marker classes
  - `PointMarker`, `PolylineMarker`, `PolygonMarker`, `SystemPointMarker`, `SelectableSystemPointMarker`, and `SystemPolylineMarker`
- `SystemPointMarker` supports current location, route location, and dropped pin points
- `SelectableSystemPointMarker` supports search pins, route start pins, and route destination pins

Note: `PolygonMarker`s do not support having a "fill color" different from the stroke color. If this effect is wanted, the developer will have to create a `PolylineMarker` and overlay it above the `PolygonMarker`

Future work will add support for more system markers such as: transit stops, shields, transit lines, and default polygons (https://github.com/mapzen/ios/issues/249)

### DIFF
--- a/src/Marker.swift
+++ b/src/Marker.swift
@@ -9,138 +9,98 @@
 import Foundation
 import TangramMap
 
-/// Enum for common marker types supported by all the house styles.
-@objc public enum MarkerType : Int {
-  case currentLocation, searchPin, routeLine
-}
-
 /// Generic marker protocol definition.
 @objc(MZGenericMarker)
 public protocol GenericMarker {
   /// The underlying Tangram marker object. Use this only for advanced cases where features of the Marker class aren't supported.
   var tgMarker: TGMarker { get }
-  /// The coordinates that the marker should be placed at on the map. Note that point, polyline, and polygon are mutually exclusive. Setting one will overwrite the other values
-  var point: TGGeoPoint { get set }
-  /// The polyline that should be displayed on the map. Note that point, polyline, and polygon are mutually exclusive. Setting one will overwrite the other values
-  var polyline: TGGeoPolyline? { get set }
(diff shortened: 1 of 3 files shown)


src/Marker.swift
```diff
@@ -235,59 +260,225 @@ public class Marker : NSObject, GenericMarker {
...
+
+  private func generateBasicStyleString() -> String {
+    return "{ style: '\(PointMarker.kPointStyle)', color: '\(backgroundColor.hexValue())', collide: false, interactive: \(interactive) }"
+  }
+
+  private func generateStyleStringWithSize() -> String {
+    if !userUpdatedSize { return generateBasicStyleString() }
+    return "{ style: '\(PointMarker.kPointStyle)', color: '\(backgroundColor.hexValue())', size: [\(size.width)px, \(size.height)px], collide: false, interactive: \(interactive) }"
+  }
+
```
msmollin: nitpick - carriage return.

src/Marker.swift
```diff
@@ -186,10 +167,63 @@ public class Marker : NSObject, GenericMarker {
...
+  override init(tgMarker tgM: TGMarker) {
+    backgroundColor = GeometricMarker.kDefaultBackgroundColor
+    interactive = GeometricMarker.kDefaultInteractive
+    super.init(tgMarker: tgM)
+  }
+
+  func updateStyleString() {
+    // override in subclass
+  }
+
```
msmollin: nitpick: whitespace

PR #258, opened 2017-04-07
Sample App Cleanup part 1

msmollin:
- Changes search tab to be in line with requirements from #224
- Refactors MapViewController to MZMapViewController
- Begins naming standardization efforts in the sample app
- Begins changing over the storyboard to use standard UIViewControllers instead of GLViewControllers

### DIFF
--- a/SampleApp/SearchPinsViewController.swift
+++ b/SampleApp/SearchPinsViewController.swift
@@ -9,16 +9,23 @@
 import Pelias
 import TangramMap
 
-class SearchPinsViewController: SampleMapViewController, UITextFieldDelegate {
+class SearchPinsViewController: SampleMapViewController, UITextFieldDelegate, AutocompleteSearchDelegate {
 
   @IBOutlet weak var searchField: UITextField!
+
+  @IBOutlet weak var displaySearch: UIButton!
+  let searchListSegueId = "searchListSegueId"
 
   override func viewDidLoad() {
     super.viewDidLoad()
 
-    try? loadStyle(.bubbleWrap)
+    try? loadStyleAsync(.bubbleWrap) { [unowned self] (style) in
+      let _ = self.showCurrentLocation(true)
+      self.showFindMeButon(true)
+    }
     
-    self.view.bringSubview(toFront: searchField)
+    view.bringSubview(toFront: searchField)
+    view.bringSubview(toFront: displaySearch)
     searchField.delegate = self
   }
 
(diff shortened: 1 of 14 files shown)


SampleApp/SearchPinsViewController.swift
```diff
@@ -50,7 +57,15 @@ class SearchPinsViewController: SampleMapViewController, UITextFieldDelegate {
...
-  // MARK:- private
-
+  func selected(_ location: PeliasMapkitAnnotation) {
+    print("Selected \(String(describing: location.title))")
+    searchField.text = location.title
+    location.setTarget(target: self, action: #selector(self.annotationClicked(annotation:)))
+    try? self.removeAnnotations()
+    try? self.add([location])
 
+    animate(toZoomLevel: 10, withDuration: 1.0)
```
sarahlensing: Nit: maybe we should use `max(10, self.zoom)` so that if the user zooms in further than level 10 or clicks the find me button, the map doesn't animate out?

ios-sdk.xcodeproj/xcshareddata/xcschemes/ios-sdk.xcscheme
```diff
@@ -74,7 +74,7 @@
       <AdditionalOptions>
       </AdditionalOptions>
       <LocationScenarioReference
-         identifier = "com.apple.dt.IDEFoundation.CurrentLocationScenarioIdentifier"
+         identifier = "New York, NY, USA"
```
sarahlensing: I'm curios how you changed this value? And what does it do?

SampleApp/SearchPinsViewController.swift
```diff
@@ -9,16 +9,23 @@
...
   @IBOutlet weak var searchField: UITextField!
 
+  @IBOutlet weak var displaySearch: UIButton!
+  let searchListSegueId = "searchListSegueId"
+
   override func viewDidLoad() {
     super.viewDidLoad()
 
-    try? loadStyle(.bubbleWrap)
+    try? loadStyleAsync(.bubbleWrap) { [unowned self] (style) in
```
sarahlensing: This is flashing a black screen, should we use `loadStyle` for now instead?

ios-sdk.xcodeproj/xcshareddata/xcschemes/ios-sdk.xcscheme
sarahlensing: Ah i see now, updated my local scheme to be based on this. Nice change

SampleApp/SearchPinsViewController.swift
msmollin: Easy enough change to make, although to be honest I thought through some of that (being zoomed in further specifically) and honestly it didn't seem that jarring to reset the view. That said, again it's a demo app so I only put minimal thought into the UX haha :D

SampleApp/SearchPinsViewController.swift
msmollin: I don't think so, honestly. Async is the "correct" approach on mobile devices. Blocking main thread is bad and all that.

@karimnaaji / @tallytalwar / @matteblair thoughts?

matteblair: The degree of "incorrectness" here depends entirely on your application :)

A synchronous scene load is blocked on I/O operations (potentially network calls) and you usually don't want to do that on a UI thread. We mostly offer this synchronous method so that you can make it asynchronous using a different mechanism if you want to.

msmollin: Do you have any idea why the async methods would show a black screen while attempting to load the scene?

matteblair: Since the render loop is, in fact, running during an async load, we're probably rendering an "empty" scene, with the default background color of `(0, 0, 0)`. There are things we could do about this.

msmollin: Talked this over with @mjcunningham - we can leave this for now in the sample app and fix it with future Tangram-es improvements.

ios-sdk.xcodeproj/xcshareddata/xcschemes/ios-sdk.xcscheme
msmollin: 👍

PR #269, opened 2017-04-17
Synchronize the podfile and podspec to match 1.0 release settings

msmollin:
### Overview
Our podspec drifted out of date from the sample app's podfile, so this brings that into sync and also updates the sample app to utilize the recently cut releases for Pelias and OnTheRoad.
### Proposed Changes
- Update Mapzen-ios-sdk.podspec to correct dependency versions
- Update sample app's podfile to use production cocoapods

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -1,36 +1,20 @@
 PODS:
-  - OnTheRoad (0.9.0)
-  - Pelias (1.0.0-beta4):
-    - Pelias/Core (= 1.0.0-beta4)
-  - Pelias/Core (1.0.0-beta4)
+  - OnTheRoad (1.0.0)
+  - Pelias (1.0.0):
+    - Pelias/Core (= 1.0.0)
+  - Pelias/Core (1.0.0)
   - Tangram-es (0.5.1)
 
 DEPENDENCIES:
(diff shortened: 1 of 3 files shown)


Podfile.lock
```diff
@@ -1,36 +1,20 @@
 PODS:
-  - OnTheRoad (0.9.0)
-  - Pelias (1.0.0-beta4):
-    - Pelias/Core (= 1.0.0-beta4)
-  - Pelias/Core (1.0.0-beta4)
+  - OnTheRoad (1.0.0)
+  - Pelias (1.0.0):
+    - Pelias/Core (= 1.0.0)
+  - Pelias/Core (1.0.0)
```
ecgreb: Why is there a separate line for `Pelias/Core` in the lock file but not other dependencies and not in any other config files?

msmollin: Pelias defines a "Core" subspec which OnTheRoad and Tangram-es do not. Your app's Podfile can define different versions for each subspec (although normally you try to avoid such things).

sarahlensing:
Changes look good, I created this issue to update our release notes so it doesnt happen again: https://github.com/mapzen/ios/issues/new

PR #272, opened 2017-04-18
House Styles Bundle

### LINKED ISSUE #225: Add copy files target from xcode project configuration
Right now some of the styles in the style switcher are missing resources. For fonts, it wouldn't be visible since we provide proper fallback when a font is missing in tangram-es, but some other resources would make the style not rendering a complete layers (walkabout style is missing the terrain layer since the spheremaps are not added). 

A simple fix to test this is to add all of the images within `scenes/styles/walkabout/images/` within a _copy files_ target in the xcode build phases to see the terrain layer appear. Maybe we could configure cocoapods so updating styles wouldn't require updating the project with new style resources that might appear.

Once styles are within the framework bundle, this wouldn't be an issue anymore.

sarahlensing:
### Overview
Creates a resource bundle for all of the house styles and loads stylesheets through it. Resolves stylesheets missing resources as well as loading issues for third party apps.

### Proposed Changes
- Updates stylesheet submodule locations to programmatic bundle (creates folder references for each stylesheet so resources are loaded dynamically)
- Removes stylesheet resource build phases, dynamically loads stylesheet dependencies
- Creates bundle extensions for accessing the mapzen sdk bundle as well as the stylesheet bundle (the SDK bundle will hold all images and any future xibs)

Closes #225

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,15 +1,15 @@
 [submodule "styles/bubble-wrap"]
-	path = styles/bubble-wrap
+	path = housestyles.bundle/bubble-wrap
 	url = https://github.com/tangrams/bubble-wrap.git
 [submodule "styles/cinnabar"]
-	path = styles/cinnabar
+	path = housestyles.bundle/cinnabar
 	url = https://github.com/tangrams/cinnabar-style-more-labels.git
 [submodule "styles/refill"]
-	path = styles/refill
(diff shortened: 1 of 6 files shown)


msmollin:
LGTM. I tested this by loading the pod into a harness project, with my existing SDK repo, and with a fresh clone. All appear to work correctly.

PR #274, opened 2017-04-18
Update podspec so it tells cocoapods to load submodules

msmollin:
### Overview
Since we changed how we load the style sheets we need to update our podspec to indicate to load that.
### Proposed Changes
- Updates the source directive to indicate the source contains submodules.


msmollin:
FYI on this - if you're pulling directly from master, your podfile directive will need to look like
`  pod "Mapzen-ios-sdk", :git => 'https://github.com/mapzen/ios.git', :branch => 'master', :submodules => true`

Otherwise it won't load submodules. This isn't a problem for actual releases though.

PR #275, opened 2017-04-18
Autocomplete crash

### LINKED ISSUE #273: Investigate Crash in sample app related to autocomplete
### Description
The latest nightlies (Circle Build 560) have a fairly reproducible crash occurring in search tab. However its not exactly the same steps each time to reproduce it, and it's not reproducible (so far) locally.

### Steps to Reproduce
1. Install nightly build from Circle CI
2. Launch app
3. Attempt to use search.

### Mapzen SDK & iOS Version
SDK: 1.0.0
Sample App: Nightly 560.

sarahlensing:
### Overview
Fixes crash in sample when executing autocomplete queries at a rate faster than the allowed queries per minute

### Proposed Changes
- Adds default query delay of 0.3 seconds to `MapzenSearch`
- Updates sample app to use `MapzenSearch`
- Adds method to determine if response is valid geojson (so that developer can ignore apiaxle errors returned until this is implemented: https://github.com/pelias/pelias-ios-sdk/issues/4)
- Fixes abrupt clearing of autocomplete result list when rate limit hit

Closes #273

### DIFF
--- a/src/PeliasMapkitExtensions.swift
+++ b/src/PeliasMapkitExtensions.swift
@@ -123,8 +123,8 @@ public extension PeliasResponse {
   public func parsedMapItems(target: UIResponder?, action: Selector?) -> [PeliasMapkitAnnotation]? {
     //TODO: This should get refactored into eventually being a real GeoJSON decoder, and split out the MapItem creation
     var mapItems = [PeliasMapkitAnnotation]()
-    if let jsonDictionary = parsedResponse?.parsedResponse {
-      guard let featuresArray = jsonDictionary.object(forKey: "features") as? [[String:AnyObject]] else {
+    if let jsonDictionary: Dictionary = parsedResponse?.parsedResponse {
+      guard let featuresArray = jsonDictionary["features"] as? [Dictionary<String, Any>] else {
         return nil
       }
       for feature in featuresArray {
--- a/src/PeliasMapkitExtensions.swift
+++ b/src/PeliasMapkitExtensions.swift
@@ -130,18 +130,20 @@ public extension PeliasResponse {
       for feature in featuresArray {
         //Address Dictionary for Placemark Creation
         let featureProperties = feature["properties"] as? [String:AnyObject]
-        var addressDictionary = [String:String]()
-        addressDictionary[PeliasIDKey] = featureProperties?["id"] as? String
-        addressDictionary[PeliasDataSourceKey] = featureProperties?["source"] as? String
+        var addressDictionary = [String:AnyObject]()
+        addressDictionary[PeliasIDKey] = featureProperties?["id"]
+        addressDictionary[PeliasDataSourceKey] = featureProperties?["source"]
 
         //Coordinate Creation
         let featureGeometry = feature["geometry"] as? [String:AnyObject]
-        let geometryPosition = featureGeometry?["coordinates"] as! [Double]
-        let coordinate = CLLocationCoordinate2DMake(geometryPosition[1], geometryPosition[0])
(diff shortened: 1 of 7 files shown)


src/PeliasMapkitExtensions.swift
```diff
@@ -124,21 +124,23 @@ public extension PeliasResponse {
...
+        return nil
+      }
       for feature in featuresArray {
         //Address Dictionary for Placemark Creation
-        let featureProperties = feature["properties"] as! [String:AnyObject]
+        let featureProperties = feature["properties"] as? [String:AnyObject]
         var addressDictionary = [String:String]()
-        addressDictionary[PeliasIDKey] = featureProperties["id"] as? String
-        addressDictionary[PeliasDataSourceKey] = featureProperties["source"] as? String
+        addressDictionary[PeliasIDKey] = featureProperties?["id"] as? String
```
msmollin: Does this work? `addressDictionary` requires a non-optional value for the key.

src/PeliasMapkitExtensions.swift
```diff
@@ -124,21 +124,23 @@ public extension PeliasResponse {
...
+        addressDictionary[PeliasDataSourceKey] = featureProperties?["source"] as? String
 
         //Coordinate Creation
-        let featureGeometry = feature["geometry"] as! [String:AnyObject]
-        let geometryPosition = featureGeometry["coordinates"] as! [Double]
+        let featureGeometry = feature["geometry"] as? [String:AnyObject]
+        let geometryPosition = featureGeometry?["coordinates"] as! [Double]
```
msmollin: should probably be `as?` ?

src/SearchResponse.swift
```diff
@@ -12,12 +12,30 @@ import Pelias
...
   let peliasResponse: PeliasResponse
 
   private lazy var internalParsedResponse: ParsedSearchResponse? =  { [unowned self] in
     guard let peliasParsedResponse = self.peliasResponse.parsedResponse else { return nil }
     return ParsedSearchResponse.init(peliasParsedResponse)
   }()
+
+  private static let validTypes: Set = ["Point", "MultiPoint", "LineString", "MultiLineString", "Polygon", "MultiPolygon", "GeometryCollection", "Feature", "FeatureCollection"]
+
+  private lazy var internalIsGeoJson: Bool = { [unowned self] in
```
msmollin: Unsure if this is the right approach.

If we don't get valid GeoJSON back - does it make sense to return a formed response object? Maybe instead return an error? Would require changing the closure parameters (honestly I thought there was an error object coming back).

src/PeliasMapkitExtensions.swift
sarahlensing: If by work you mean doesn't crash, then yes :) The way it works is that the key `PeliasIDKey` is not added to `addressDictionary` if there is not an "id" in the `featureProperties`. I think this is expected...but looking at `PeliasMapKitAnnotation`, it looks like we cast it to `[String:AnyObject]` so we can probably update addressDictionary and remove the string casting...will do that now

src/PeliasMapkitExtensions.swift
sarahlensing: Agreed, nice catch

src/SearchResponse.swift
sarahlensing: I agree with returning an error in that case...I had pushed [this change](https://github.com/pelias/pelias-ios-sdk/pull/75) which will return errors if we dont get valid geojson. When it merges we can update pelias...i will remove this variable since it shouldnt be needed, the presence of a parsed response will mean we got valid geojson

PR #278, opened 2017-04-19
Add crash reporting for sample app

### LINKED ISSUE #276: [discuss] Add crash reporting
We should add crash reporting into the sample app. 

Using HockeyApp seems like a good first place to start since we are already using it for distribution: https://support.hockeyapp.net/kb/client-integration-ios-mac-os-x-tvos/hockeyapp-for-ios#crashreporting. With hockeyapp we can allow auto-sending of reports or user confirmation.

sarahlensing:
### Overview
Adds HockeyApp crash reporting for SDK sample app

### Proposed Changes
- Refactors environment variable logic into separate metod
- Updates circle.yml to inject key when building ipa
- Added new HOCKEY_APP_ID into circle's environment variables

Closes #276

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -9,6 +9,7 @@ end
 
 target "ios-sdk" do
   shared_pods
+  pod "HockeySDK", :subspecs => ['CrashOnlyLib']
 end
 
 target "ios-sdkTests" do
--- a/Podfile
+++ b/Podfile
@@ -9,7 +9,7 @@ end
 
 target "ios-sdk" do
   shared_pods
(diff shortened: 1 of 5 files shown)


Podfile
```diff
@@ -9,6 +9,7 @@ end
 
 target "ios-sdk" do
   shared_pods
+  pod "HockeySDK", :subspecs => ['CrashOnlyLib']
```
msmollin: Let's lock this to a specific version series like `pod "HockeySDK", ~> "4.1", :subspecs => ['CrashOnlyLib']` 

Would rather not get surprised by a major version release update.

PR #284, opened 2017-04-21
Update pelias

sarahlensing:
### Overview
Pulls in new version of Pelias which fixes bug where common search query params were not being appended to requests. This meant that when MapzenSearch used the client-side rate limiter, the api key was not appended to requests

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -2,7 +2,7 @@ platform :ios, '9.3'
 use_frameworks!
 
 def shared_pods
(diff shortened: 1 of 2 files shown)


msmollin:
Note that only happened on Autocomplete as its currently the only thing the Pelias SDK rate limits, which made this all the much more fun to track down.

msmollin:
👍

Issue #286, opened 2017-04-24
Corrupted text on iPhone 6


bcamper:
### Description
I'm seeing a couple variants of corrupted or missing text.

Random characters / corrupt label:
![img_5417](https://cloud.githubusercontent.com/assets/16733/25344394/fabbb6d8-28df-11e7-8b31-761eba00a960.PNG)

Missing labels:
![img_5420](https://cloud.githubusercontent.com/assets/16733/25344403/fe899fd2-28df-11e7-86c2-a6079bb37d87.PNG)

### Steps to Reproduce
- Opened app, zoomed to Central Park, noticed corrupt labels
- Zoomed out, noticed missing text on shields
- Cannot reproduce every time I use the app, but reproduced missing shield text on second or third attempt (have only seen that specific type of corrupt label text in first screenshot the one time).

### Mapzen SDK & iOS Version
SDK 0.4.0
iOS 10.2.1
iPhone 6

meetar:
FTR I can't reproduce on iPhone 5s with iOS 10.3.1.

msmollin:
Been able to reproduce this twice, but only once per device. Subsequent runs and rebuilds and app deletions / reinstalls have been fine.

iPad Mini 1 running iOS 9.3.5:
![screen shot 2017-04-24 at 12 08 41 pm](https://cloud.githubusercontent.com/assets/1163880/25347121/d66610aa-28e7-11e7-9792-b398438b7744.png)

iPad Mini 2 Running iOS 10.2.1:
![screen shot 2017-04-24 at 12 08 41 pm](https://cloud.githubusercontent.com/assets/1163880/25347131/dcc68416-28e7-11e7-9e5a-fde2014d92c2.png)

msmollin:
Have not seen the font issue that @bcamper originally mentioned on any device.

msmollin:
Just to follow up here:
@tallytalwar and @hjanetzek discovered an issue inside the alfons project that Tangram-es utilizes to load in and store font-faces. This issue is being fixed upstream in alfons which then will result in needing updating inside Tangram-es. 

Additionally the Tangram-es iOS framework will introduce changes to limit the number of system font families it attempts to load when attempting to render certain glyphs.

[...]

PR #294, opened 2017-04-26
Remove fonts from Submodules and use system fonts for 1.0.0 release

msmollin:
### Overview
This change removes the stylesheet bundled fonts and relies upon system font loading for the 1.0.0 release until fixes can be merged from Tangram-es to properly address #286


tallytalwar:
Since this change will affect the look of the text in the bundled scenes, we decided that its probably best for @nvkelso to have a look at the rendered scene with the change.

Since @matteblair is visiting SF, he will be confirming with our carto team and coordinating the change. Once I get a 👍 from the carto team will merge this.

msmollin:
@tallytalwar OK I listed a more detailed explanation in #286 if you want to add anymore details / provide corrections.

matteblair:
Ok I've checked in with the carto team and everyone agrees that it's better to ship with only the iOS system fonts at launch than to ship with garbled text.

@msmollin Once you've confirmed that this prevents the issue from surfacing in extended use, I think we're good to merge this.

msmollin:
Roger and thanks for talking to them. I have all the devices I replicated this on at home so I'll exercise this tomorrow and report back.

tallytalwar:
I checked this on 10.0.x iphone and 10.3.1 iphone we have at work today. Looks good.

msmollin:
Checked this on iOS 9.3.5 iPad Mini 1 by starting in Baghdad, then going to NYC central park, then Seoul, then downtown Bejing, and then Moscow, all of which I got to load different glyph styles. Upon going back to NYC there was still no issues. The same series of events also didn't showcase anything on my 10.2.1 iPad Air, so this seems to have resolved the issue.

PR #311, opened 2017-05-10
Update documentation to align with parameters

### LINKED ISSUE #298: Fix warnings in Objective-C documentation generation
### Description
When outputting the documentation from the swift code into objective-c, we get warning generated because of mismatched labels (even though swift is entirely happy with it).

```  /**
   Loads a map style synchronously on the main thread. Use the async methods instead of these in production apps.

   - parameter style: The map style to load.
   - parameter locale: The locale to use for the map's language.
   - parameter sceneUpdates: The scene updates to make while loading the map style.
   - throws: A MZError `apiKeyNotSet` error if an API Key has not been sent on the MapzenManager class.
   */
  open func loadStyle(_ style: MapStyle, locale l: Locale, sceneUpdates: [TGSceneUpdate]) throws {
```

results in Obj-c projects: 
screen shot 2017-04-27 at 11 54 06 am

### Mapzen SDK & iOS Version
SDK: 0.4.0

msmollin:
### Overview
This resolves the warnings in objective-c related to documentation not aligning with parameter names.
### Proposed Changes
We landed on just resolving the documentation issues rather than changing public-facing API since we're post-1.0.0. However we should come back in 1.1.0 and deprecate these functions and properly rename them. 

Fixes #298

### DIFF
--- a/MapzenSDK/Marker.swift
+++ b/MapzenSDK/Marker.swift
@@ -42,7 +42,7 @@ public protocol GenericPointMarker: GenericMarker {
 
    - parameter coordinates: Coordinates to animate the marker to
    - parameter seconds: Duration in seconds of the animation.
-   - parameter easeType: Easing to use for animation.
+   - parameter ease: Easing to use for animation.
    */
   func setPointEased(_ coordinates: TGGeoPoint, seconds: Float, easeType ease: TGEaseType) -> Bool
 
@@ -57,7 +57,7 @@ public protocol GenericPointIconMarker: GenericGeometricMarker, GenericPointMark
   /**
    Initializes a marker with a given size.
 
-   - parameter size: The size the marker should be.
+   - parameter s: The size the marker should be.
    */
   init(size s: CGSize)
 }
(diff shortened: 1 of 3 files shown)


MapzenSDK/Marker.swift
```diff
@@ -256,7 +256,7 @@ public class PointMarker : GeometricMarker, GenericPointIconMarker {
   /**
    Initializes a marker with a given size.
 
-   - parameter size: The size the marker should be.
+   - parameter s: The size the marker should be.
```
karimnaaji: Would there be some nicer code style to adopt concerning this without shading the type or private class names? Not sure what's swifty around this kind of things but one letter parameters for public interface seem to be ambiguous.

msmollin: Indeed there is - it's just the `s` is now public facing as of 1.0 so it's harder to just change that. The plan is to deprecate these methods in 1.1.0 and then remove in 2.0.0 to clean that up.

PR #325, opened 2017-05-18
Add OSM Default language choice to the language selector

msmollin:
### Overview
This enables the sample map to be viewed using whatever the default language is for the region, which can help when debugging font issues.
### Proposed Changes
Add an "OSM Default" language setting, and set an update to remove the language scene setting.

### DIFF
--- a/SampleApp/DemoMapViewController.swift
+++ b/SampleApp/DemoMapViewController.swift
@@ -7,6 +7,7 @@
 //
 
 import UIKit
+import TangramMap
 import Mapzen_ios_sdk
 
 class DemoMapViewController:  SampleMapViewController, MapMarkerSelectDelegate {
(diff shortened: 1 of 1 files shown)


SampleApp/DemoMapViewController.swift
```diff
@@ -97,9 +98,18 @@ class DemoMapViewController:  SampleMapViewController, MapMarkerSelectDelegate {
       "Spanish": "es_ES",
       "Korean": "ko_KR",
       "Italian": "it_IT",
+      "OSM Default": "none",
       ]
     let actionSheet = UIAlertController.init(title: "Map Language", message: "Choose a language", preferredStyle: .actionSheet)
     for (actionTitle, languageIdentifier) in languageIdByActionSheetTitle {
+      if (languageIdentifier == "none") {
+        actionSheet.addAction(UIAlertAction.init(title: actionTitle, style: .default, handler: { [unowned self] (action) in
+          let update = TGSceneUpdate(path: "global.ux_language", value: "")
```
karimnaaji: Isn't that the job of the SDK method `updateLocale`? We could rather make it accept an empty locale or a `nil` locale and fill up the `return` statement [here](https://github.com/mapzen/ios/blob/d0e60fc0207ba25853297770b3809c4412cdda84/MapzenSDK/MZMapViewController.swift#L622) to update the global ux language to en empty string, that would give a good default behavior.

karimnaaji: Hmm actually the default value is already an empty string so it's fine.

PR #336, opened 2017-06-19
Update placeholder text, fix link

rmglennon:
Update the placeholder text as part of https://github.com/mapzen/documentation/issues/310.

I also deleted some text that describes how to use the placeholder text (like replace it with your real key) because the key is already replaced when signed in, and the pop-up over the text will be the call to action for users who are not signed in.

Also fixed link formatting and it's --> its.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -1,17 +1,17 @@
 # Getting started
 
 ## 1. Sign up for an API key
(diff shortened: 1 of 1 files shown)


rmglennon:
This is ready for review and merge now. Testing looks OK for this section.

If you want to preview, you can go to https://dev.mapzen.com/documentation/ios/getting-started/

When signed in, the placeholder text swaps to the first of your API keys. If you are not signed in or have no API keys, you still see your-mapzen-api-key.

rmglennon:
After we finish off this PR, I'll be able to close https://github.com/mapzen/documentation/issues/310 for the placeholder text updates.

There is some work this quarter to finish up the code to make the swapping happen on the public website, as we found some bugs in other sections.

Issue #248, opened 2017-03-29
When Tangram 0.5.2 comes out, add back in error handing for markers and scenes


msmollin:

karimnaaji:
Debug preview build available here: https://github.com/tangrams/ios-framework/tree/tangram-0.7-debug-preview

PR #342, opened 2017-06-22
Upgrade Tangram-ES to 0.7

### LINKED ISSUE #327: Upgrade to Tangram-es 0.7
Will be a bit of a refactor in the marker interface section but overall should be mostly drop-in.

msmollin:
### Overview
This PR upgrades Tangram-es to the latest production stable version. It includes some major refactoring, some method deprecations, and some bug fixes and performance improvements. 

**Note that this includes a minor breaking change where we have to make the tgMarker property across the marker protocols optional now. This shouldn't affect most people but it should be called out in release notes.**

### Proposed Changes
- Updates the podspec to pull in latest changes
- Update build settings to satisfy xcode
- Alter unit tests and main code to satisfy API and functionality changes 
- SDK markers are now decoupled from ES markers. This has some good functionality benefits but means we need to do additional checks before adding them to the map to avoid overwriting changes that may have happens to the TGMarker
- Request that GLKit render the map at 60 FPS which makes for a much smoother and nicer mapping experience.
- Minor bug fix to actually remove all the markers from our internal storage. 

Fixes #327


MapzenSDK/MZMapViewController.swift
```diff
@@ -487,8 +487,10 @@ open class MZMapViewController: UIViewController, LocationManagerDelegate {
    - parameter marker: The marker to add to map.
    */
   open func addMarker(_ marker: GenericMarker) {
-    currentMarkers[marker.tgMarker] = marker
-    marker.tgMarker.map = tgViewController
+    if marker.tgMarker == nil {
+      marker.tgMarker = tgViewController.markerAdd()
+    }
+    currentMarkers[marker.tgMarker!] = marker
```
karimnaaji: Could this be useful or is this aiming for something different? https://github.com/tangrams/tangram-es/blob/master/platforms/ios/src/TangramMap/TGMapViewController.h#L319

karimnaaji: Looking more, it seems that a nice alternative would be to store the SDK marker as a user data of the Tangram marker, that the association can be easily retrieved without additional container.

https://github.com/tangrams/tangram-es/blob/master/platforms/ios/src/TangramMap/TGMarker.h#L103

I think it reminds me of some conversation we had on a previous PR relating to that.

msmollin: It's a slightly different use case. With this change we're effectively decoupling SDK markers from Tangram-es markers, meaning we no longer require a Tangram-es marker be created at initialization time of the SDK marker class.  We're also allowing users to set their own underlying TGMarkers for their SDK markers if they want (it honestly makes it easier on us to implement as well).

The side effect of all that is we need to check we don't override previous markers when adding the SDK markers to the map.

msmollin: Ahhh slow github is slow - yeah the user data object is a good idea. Lemme think on how that might change the interaction model at the Mapzen layer.

MapzenSDK/Marker.swift
```diff
@@ -411,15 +457,23 @@ public class SystemPointMarker : Marker, GenericSystemPointMarker {
...
    - parameter seconds: Duration in seconds of the animation.
    - parameter ease: Easing to use for animation.
+   - returns: True if successful, false if error or marker has not been added to the map
    */
   public func setPointEased(_ coordinates: TGGeoPoint, seconds: Float, easeType ease: TGEaseType) -> Bool {
-    return tgMarker.setPointEased(coordinates, seconds: seconds, easeType: ease)
+    guard let marker = tgMarker else { return false }
+    marker.pointEased(coordinates, seconds: seconds, easeType: ease)
+    do {
+      try marker.getError()
```
karimnaaji: The error is only checked on `setPointEased`, I assume that the inconvenience of Swift regarding properties that throw is a potential blocker, but the implementation might leave error incoming from previous usage of the marker, which would be misleading when using `setPointEased`. What seems correct is to check on every modification of the marker or check nowhere at all (while potentially wrapping `getError` to allow its usage externally).

msmollin: Hmmm yeah you're right. Probably should just pull this then and go back through a re-add it when we tackle https://github.com/mapzen/ios/issues/248

MapzenSDK/Marker.swift
```diff
@@ -234,12 +245,20 @@ public class PointMarker : GeometricMarker, GenericPointIconMarker {
    - parameter ease: Easing to use for animation.
    */
   public func setPointEased(_ coordinates: TGGeoPoint, seconds: Float, easeType ease: TGEaseType) -> Bool {
-    return tgMarker.setPointEased(coordinates, seconds: seconds, easeType: ease)
+    tgMarker?.pointEased(coordinates, seconds: seconds, easeType: ease)
```
sarahlensing: ~Do we need to call `getError` before `pointEased` to avoid returning an error for a previous action taken on `tgMarker`? I'm wondering if we have an existing error on the `tgMarker` and then call `pointEased` which completes without an error, will we erroneously return `false` here?~ I just saw @karimnaaji's comment, looks like you already plan to handle this

MapzenSDK/MZMapViewController.swift
msmollin: Ok so @sarahlensing and I discussed it further and I think we wanna go down the route of using the user data object on TGMarker. However, a void* there imported into Swift is difficult to use. Do you think it would be much effort to convert it to a nullable `NSObject` ?

karimnaaji: `NSObject` seems to narrow down the flexibility quite a lot, we should probably use`id` there?

msmollin: `id` could work. That at least converts to `AnyObject` in Swift which I can cast through easier.

msmollin: The other places I see in Foundation that have this type of construct (`NSNotification` for example) flat out use `NSDictionary` which is even more constraining.

karimnaaji: I actually like the `NSDictionary` idea, it would let you store different kind of data without having to create your own data structure holding it. Feel free to open an issue in ES with the different options we have.

SampleApp/DemoMapViewController.swift
```diff
@@ -34,10 +34,11 @@ class DemoMapViewController:  SampleMapViewController, MapMarkerSelectDelegate {
     setupStyleObservance()
     markerSelectDelegate = self
     let appDelegate = UIApplication.shared.delegate as! AppDelegate
-    try? loadStyleAsync(appDelegate.selectedMapStyle) { [unowned self] (style) in
-      self.styleLoaded = true
-      let _ = self.showCurrentLocation(true)
-      self.showFindMeButon(true)
+    try? loadStyleAsync(appDelegate.selectedMapStyle) { [weak self] (style) in
+      guard let unwrappedSelf = self else { print("Self is nil"); return }
```
karimnaaji: nit: Might be useful to have a more descriptive message here, and maybe a description of how this can happen.

msmollin: Hah that is debug code - had a separate issue I was trying to figure out and forgot to remove that. Will do now.

PR #352, opened 2017-08-12
Upgrade to Tangram-es 0.8 and update internal code for API changes Part 1

msmollin:
### Overview
This updates the SDK to use Tangram-es 0.8 and includes the necessary refactors along with some cleanup associated with it.

### Proposed Changes
- Upgrades to Tangram-es 0.8
- Removes memory management from MZMapViewController
- Deprecates a number of methods in preparation for eventual removal.

Note that this PR lacks marker state restoration. That will be coming in a future PR.


MapzenSDK/MZMapViewController.swift
```diff
@@ -1177,21 +1071,29 @@ open class MZMapViewController: UIViewController, LocationManagerDelegate {
...
+    sceneLoadCallback = nil
+  }
+
+  func restoreBuiltInMarkers() {
+    //Handle built in marker restoration
+    if shouldShowCurrentLocation && currentLocationGem != nil {
+      let lastPosition = currentLocationGem?.point
+      currentLocationGem = nil
+      _ = showCurrentLocation(true)
+      currentLocationGem?.point = lastPosition!
```
sarahlensing: Any reason not to `guard let` on line 1091 and avoid force unwrapping here?

MapzenSDK/MZMapViewController.swift
```diff
@@ -1177,21 +1071,29 @@ open class MZMapViewController: UIViewController, LocationManagerDelegate {
...
+    //TODO: For 2.0 we should pass the Error along in the callback block.
+    if sceneID != latestSceneId {
       return
     }
-    // if we have path components, grab the last two (ie. bubble-wrap & bubble-wrap-style-more-labels.yaml), strip ".yaml" and check for existence in styles map
-    let sceneStyle = (pathComponents[pathComponents.count-2] + "/" + pathComponents.last!).replacingOccurrences(of: ".yaml", with: "")
-    guard let style = styles[sceneStyle] else {
-      onStyleLoadedClosure = nil
-      return
+    restoreBuiltInMarkers()
```
sarahlensing: We might want to move this to the top of the method and always restore built in markers so that other map methods like enabling transit overlays etc. don't cause built in markers to disappear.

MapzenSDKTests/MapViewControllerTests.swift
```diff
@@ -629,7 +627,7 @@ class MapViewControllerTests: XCTestCase {
     var loaded = false
     let styleLoaded : MZMapViewController.OnStyleLoaded = { (style) in loaded = true }
     try? controller.loadStyleAsync(.bubbleWrap, onStyleLoaded: styleLoaded)
-    controller.mapView(controller.tgViewController, didLoadSceneAsync: "/test/container/bubble-wrap/bubble-wrap-style-more-labels.yaml")
+    controller.mapView(controller.tgViewController, didLoadScene: 0, withError: nil)
```
sarahlensing: Should this be `didLoadSceneAsync` ?

MapzenSDK/MZMapViewController.swift
msmollin: My only concern there is we don't want to be restoring markers if we have another scene load coming through, which is why this happens underneath the `sceneId` check. We'd basically be wasting cycles. I can test the transit overlay scenario more but I don't recall running into that problem.

MapzenSDK/MZMapViewController.swift
msmollin: Two issues with a guard statement:
- Guards require an exit of current scope when they fail, so the rest of the function won't execute in the event a guard check fails. I conceived this function doing more than just restoring the current location gem, so I was trying to set a pattern to not just "exit". Obviously this is not actually occurring right now, but it should mean less refactoring in the future. 
- Brevity - we do the check in the `if` statement to make sure `currentLocationGem` is not nil, so it's relatively safe to force unwrap. `point` itself is a required property in the protocol declaration, so the optional state is inherited from the location gem. 

However, thinking on this a bit, I think I could swap this out to using a `if let` and make everyone happy. Lemme give that a whirl.

MapzenSDKTests/MapViewControllerTests.swift
msmollin: Nada - the callback from tangram-es is now `didLoadScene`.

MapzenSDK/MZMapViewController.swift
tallytalwar: Sorry for jumping in late, but should sdk provide interface for the user to decide if they want to retain the markers from a previous scene load when a new scene load is requested?

PR #353, opened 2017-08-14
Update the Sample App to correctly handle empty string API keys

msmollin:
### Overview
The info dictionary returns an empty string (rather than nil) for our API key by default in the objective-c sample. This causes an issue when trying to set environment variables for the API key to be injected. This PR fixes that by adding additional checks for that case.
### Proposed Changes
- Add empty string check to API key checking code.

### DIFF
--- a/SampleApp-Objc/AppDelegate.m
+++ b/SampleApp-Objc/AppDelegate.m
@@ -24,7 +24,7 @@
 }
 
 - (NSString *)getEnvironmentVariableForKey:(NSString *)key {
-  if (NSBundle.mainBundle.infoDictionary[key]) {
+  if (NSBundle.mainBundle.infoDictionary[key] && (NSString*)NSBundle.mainBundle.infoDictionary[key] != [NSString stringWithFormat:@""]) {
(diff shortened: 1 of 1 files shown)


SampleApp-Objc/AppDelegate.m
```diff
@@ -24,7 +24,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 }
 
 - (NSString *)getEnvironmentVariableForKey:(NSString *)key {
-  if (NSBundle.mainBundle.infoDictionary[key]) {
+  if (NSBundle.mainBundle.infoDictionary[key] && (NSString*)NSBundle.mainBundle.infoDictionary[key] != [NSString stringWithFormat:@""]) {
```
matteblair: Is pointer comparison what we want here? Maybe checking `length == 0` is more straightforward.

msmollin: Oh bother. That's my Swift showing - swift overloads != and == for strings to do proper comparison.

Yeah either length check or using `isEqualToString` although i think a length check here is more readable.

PR #356, opened 2017-08-23
Update SDK and Sample App to take into account controlling walking overlays

### LINKED ISSUE #347: bike map mode should disable walking network
### Description

The new v1.1.0 alpha build includes an option to show the bike map overlay, great!

But looks like the walking networks are still displayed. The recommendation in the cartography documentation is to disable those (like disabling transit).

### Steps to Reproduce

Go to Chicago, see both bike lanes and walking networks.

http://tangrams.github.io/walkabout-style/#16/41.8860/-87.6213

App shows both bike lanes and walking routes at same time (should only show bike):

img_8077

Desired view from Github.io link above:

screen shot 2017-07-13 at 16 31 02

Walking routes only view (Walkabout default):

screen shot 2017-07-13 at 16 16 43

msmollin:
### Overview
The current SDK and Sample App provide no access to managing the walking path overlays. This conflicts with our Cartography docs which recommends developers enable / disable walking paths when disabling / enabling bike paths (respectively). 
### Proposed Changes
- Add new computed property to manage walk path overlay being enabled
- Tie into the Sample app and enable the sample app to enable / disable walking and biking paths at the same time.
- Unit tests added for new properties. 

Fixes #347

### DIFF
--- a/MapzenSDK/MZMapViewController.swift
+++ b/MapzenSDK/MZMapViewController.swift
@@ -25,6 +25,7 @@ public struct StateReclaimer {
 public struct GlobalStyleVars {
   public static let transitOverlay = "global.sdk_transit_overlay"
   public static let bikeOverlay = "global.sdk_bike_overlay"
+  public static let pathOverlay = "global.sdk_path_overlay"
   public static let apiKey = "global.sdk_mapzen_api_key"
   public static let uxLanguage = "global.ux_language"
 }
@@ -234,6 +235,7 @@ open class MZMapViewController: UIViewController, LocationManagerDelegate {
   var currentStyle: MapStyle = .bubbleWrap
   var transitOverlayIsShowing = false
   var bikeOverlayIsShowing = false
+  var walkingOverlayIsShowing = false
 
   /// The camera type we want to use. Defaults to whatever is set in the style sheet.
   open var cameraType: TGCameraType {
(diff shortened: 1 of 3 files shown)


MapzenSDK/MZMapViewController.swift
```diff
@@ -623,6 +641,9 @@ open class MZMapViewController: UIViewController, LocationManagerDelegate {
     onStyleLoadedClosure = onStyleLoaded
     guard let sceneFile = styles.keyForValue(value: style) else { return }
     currentStyle = style
+    if style == .walkabout {
+      showWalkingPathOverlay = true
```
karimnaaji: Same issue here.

MapzenSDK/MZMapViewController.swift
```diff
@@ -582,6 +597,9 @@ open class MZMapViewController: UIViewController, LocationManagerDelegate {
     locale = l
     guard let sceneFile = styles.keyForValue(value: style) else { return }
     currentStyle = style
+    if style == .walkabout {
+      showWalkingPathOverlay = true
```
karimnaaji: We might not need to go through an `applySceneUpdate` and the whole process of updating the scene through Tangram here, we could set `walkingOverlayIsShowing = true` instead so `allSceneUpdates` picks it up and append the update.

MapzenSDK/MZMapViewController.swift
```diff
@@ -643,6 +664,9 @@ open class MZMapViewController: UIViewController, LocationManagerDelegate {
     onStyleLoadedClosure = onStyleLoaded
     guard let sceneFile = styles.keyForValue(value: style) else { return }
     currentStyle = style
+    if style == .walkabout {
+      showWalkingPathOverlay = true
```
karimnaaji: Same here.

MapzenSDK/MZMapViewController.swift
msmollin: Good catch / optimization. I'll give it try.

PR #358, opened 2017-09-05
Tangram 0.8 Upgrade Part 2 - Enable Marker Restoration

smolls:
### Overview
Tangram 0.8 removes built-in support of marker restoration on scene change / update. This re-enables that for Mapzen SDK users.

Includes minor refactors / code organization.

### Proposed Changes
- Enable marker restoration for current location gem as well as user markers (routes are included in this since they're added by the user).
- New `internal` facing function for restoring markers and moving individual functions into more private scope.
- Refactor single extension into multiple extensions for individual delegate conformance.
- TGGeoPoint `equatable` conformance to make testing easier.

### DIFF
--- a/MapzenSDK/MZMapViewController.swift
+++ b/MapzenSDK/MZMapViewController.swift
@@ -1,4 +1,4 @@
-  //
+//
 //  MZMapViewController.swift
 //  ios-sdk
 //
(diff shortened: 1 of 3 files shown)


msmollin:
Wrong GH account. Sigh.

PR #364, opened 2017-09-27
Add origin tracking and cocoapods version injection

### LINKED ISSUE #329: Override User-agent to enable tracking of Mobile-SDK usage
Per discussion that occurred here: https://github.com/mapzen/ios/issues/310 we should override the user agent to provide details for tracking.

The default user agent that NSURLSession default configuration sends up looks like `$(CFBundleName)/$(CFBundleVersion) CFNetwork/808.3 Darwin/16.3.0` which is pretty much useless for our case. 

What we want is something like
`User-Agent: <sdk_name>;<sdk_version>;<os_version>`

Thoughts on formatting @sarahlensing ?

msmollin:
### Overview
Add in origin tracking for metrics collection and device usage statistics. 
### Proposed Changes
- Enable User-agent overrides to properly track outbound requests
- Injection SDK version from cocoapods using swift script process

Fixes #329


MapzenSDK/MZMapViewController.swift
```diff
@@ -910,6 +910,9 @@ open class MZMapViewController: UIViewController, LocationManagerDelegate {
 
   override open func viewDidLoad() {
     super.viewDidLoad()
+    let httpHandler = TGHttpHandler()
+    httpHandler.httpAdditionalHeaders = MapzenManager.sharedManager.httpHeaders() as! NSMutableDictionary
```
sarahlensing: This crashes because swift dictionaries aren't bridged to mutable dictionaries. You need something like `httpHandler.httpAdditionalHeaders = NSMutableDictionary(dictionary: MapzenManager.sharedManager.httpHeaders())`

msmollin: Oh? Doesn't crash for me. Hmmmm. Will recheck that.

msmollin: I've retested this inside, outside, and backwards and I cannot get it to crash. Stepping over it shows values getting set correctly on all objects involved. 

There is bridging for Dictionary types documented here: https://developer.apple.com/library/content/documentation/Swift/Conceptual/BuildingCocoaApps/WorkingWithCocoaDataTypes.html

However it's not performing 100% the way I would expect it to based on reading that documentation; Tangram objective-c code imported into Swift should have NSMutableDictionary types replaced with straight swift Dictionary types. However that might only happen reliably if the objective-c code in use is making use of lightweight typing.

That said, I will rework this a bit to avoid the forced downcast using your suggestion which does work.

sarahlensing: Thats odd, its very reproducible without the fix. Running the sample app results in a crash every time:

screenshot
 
Anyway, code looks good now. After a rebase its good to go.

PR #366, opened 2017-10-03
Styles, Themes, and 512 Tiles support

msmollin:
### Overview
This change overhauls how the SDK interacts with style sheets. Previously, we utilized a minimal but functional interface for loading stylesheets using a simple enumeration and correlative array. This works well, but it has two major downsides:

1) It allows leakage of underlying framework and stylesheet changes because it is so minimal. For example, the deprecation and removal of the `Zinc` standalone stylesheet would effectively break public-facing API if we continued to stick with this method.
2) An enumeration we control means that support custom or remotely-deployed stylesheets would be a fairly large hack to enable either of those major requested features in the existing API.

This change thus introduces a `StyleSheet` and `Theme` protocol, which allows us to express the current style sheets Mapzen provides, and the themes attached to them. Analogous to this are the introduction of loading functions for these, as well as refactoring across the sample app to support these.
### Proposed Changes
- Introduce StyleSheet And Theme Protocols
- Implement existing stylesheets using new protocols 
- Add functionality to MZMapViewController to allow loading of objects conforming to the protocols
- Revamp sample app to expose new style sheet functionality
- Remove `Zinc` stylesheet and implement new stylesheet for Zinc subclassed off Refill with theme set to `Zinc`
- Upgrade to Tangram-es 0.8.1 to support all of this.
- Upgrade Pelias to 1.0.2 for some minor fixes.
- Upgrade to use Xcode 8.3.3 on circle

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,13 +3,10 @@
 	url = https://github.com/tangrams/bubble-wrap.git
 [submodule "styles/cinnabar"]
 	path = housestyles.bundle/cinnabar
-	url = https://github.com/tangrams/cinnabar-style-more-labels.git
+	url = https://github.com/tangrams/cinnabar-style.git
 [submodule "styles/refill"]
 	path = housestyles.bundle/refill
-	url = https://github.com/tangrams/refill-style-more-labels.git
+	url = https://github.com/tangrams/refill-style.git
 [submodule "styles/walkabout"]
 	path = housestyles.bundle/walkabout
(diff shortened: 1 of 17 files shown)


msmollin:
@karimnaaji / @sarahlensing I have an outstanding support request to fix the circle ci builds for these. If you want to test things manually, when you pull the branch down make sure you run `git submodule sync` to get the new changes, and then `pod install` so you get new tangram and pelias.

msmollin:
Builds  are fixed!

PR #369, opened 2017-10-16
Overhaul stylesheet system to make compatible with Obj-c

msmollin:
### Overview
The original implementation of Stylesheets and Themes, while very nice from a Swift perspective, made it unavailable in objective-c; protocol extensions were to blame as they are unavailable to the objective-c runtime. This required a rather extensive refactor and also yielded an additional ticket to go through and check the rest of our swift-based protocols to make sure everything works correctly.
### Proposed Changes
- Remove `Theme` as a separate protocol and merge into `StyleSheet` protocol
- Create `BaseStyle` class that includes the protocol extension code
- Remove protocol extensions
- Move all styles to be subclasses of `BaseStyle` and refactor accordingly
- Add some additional Objective-c specific tests. 
- Refactor tests to take into account all the refactoring :)


MapzenSDK/Themes.swift
```diff
@@ -10,149 +10,224 @@ import Foundation
...
+@objc public protocol StyleSheet : class {
+
+  @objc var fileLocation: URL? { get }
+  @objc var remoteFileLocation: URL? { get }
+  @objc var styleSheetRoot: String { get }
+  @objc var houseStylesRoot: String { get }
+  @objc var styleSheetFileName: String { get }
+  @objc var importString: String { get }
+  @objc var relativePath: String { get }
+  @objc var mapStyle: MapStyle { get set }
```
sarahlensing: Do we need this enum still? Or do you want to deprecate it and remove it in a later release?

If we keep it, should it be a get-only property so that something like:
```swift
let bubblewrap = BubbleWrapStyle()
bubblewrap.mapStyle = .cinnabar
```
isn't possible

MapzenSDK/Themes.swift
```diff
@@ -10,149 +10,224 @@ import Foundation
...
+  @objc override open var styleSheetRoot: String {
+    get {
+      return "cinnabar/"
+    }
+  }
 
-  open var detailLevel: Int = 0 // Not used for Cinnabar
+  @objc override open var availableLabelLevels: Int {
+    get {
+      return 12
```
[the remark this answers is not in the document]
msmollin: We would *love* that - have just been told in the past that was really hard :)

MapzenSDK/Themes.swift
msmollin: Duh that makes sense.

PR #385, opened 2017-11-01
Add Background Location support & fix background GPU usage

### LINKED ISSUE #92: Background Location Update Support
Need to figure out where we draw the line with our internal handling of backgrounding and where the implementing developer needs to be involved.

### LINKED ISSUE #378: gpus_ReturnNotPermittedKillClient error occuring in iOS background-aware applications
We received some feedback in the Mapzen iOS SDK causing crashes for applications going into the background. This seems to stem from GLKitViewController not pausing its rendering (like I thought it did) when going into and out of background. On reflection this does make some sense as it would be silly for Apple to assume it knows the state of the render pipeline.

To resolve this, we should follow the guidelines outlined in https://developer.apple.com/library/content/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/ImplementingaMultitasking-awareOpenGLESApplication/ImplementingaMultitasking-awareOpenGLESApplication.html#//apple_ref/doc/uid/TP40008793-CH5-SW1

https://github.com/tangrams/tangram-es/issues/1708 has been opened in the upstream Tangram-es repo as that is probably the correct place to implement this. However, we can probably implement something in the iOS-SDK sooner if necessary.

### LINKED ISSUE #390: Location update when App in background
Hi guys,
I am trying to make my app update my current location at all times but unfortunately it does not work: after locking my screen or after having my app in the background, the little arrow on the top right corner disappears.
I have added the "Privacy - Location Always Usage" and the "Privacy - Location when in Use" in the info.plist but it does not change anything. 
I also set the app on my phone to "Always" in the Privacy/Location Services Settings.
Is there something else that I missed?

msmollin:
### Overview
This fixes https://github.com/mapzen/ios/issues/378 and additionally adds support for background location mode updates in the main location manager. 
### Proposed Changes
- Modifies LocationManager to support enabling background location updates
- Adds background mode awareness so we no longer try to call animation code while in background
- Adds testing around these implementations.
- Fixes #378 
- Fixes #390 
- Fixes #92


MapzenSDK/MZMapViewController.swift
```diff
@@ -976,10 +986,31 @@ open class MZMapViewController: UIViewController, LocationManagerDelegate {
...
+  }
+
+  deinit {
+    NotificationCenter.default.removeObserver(self)
+  }
+
+  //MARK: - Application Lifecycle Observance
+  func observeLifecycleNotifications(notificationCenter: NotificationCenterProtocol = NotificationCenter.default) {
+    notificationCenter.addObserver(self, selector: #selector(applicationWillResignActive), name: .UIApplicationWillResignActive, object: nil)
+    notificationCenter.addObserver(self, selector: #selector(applicationWillEnterForeground), name: .UIApplicationWillEnterForeground, object: nil)
```
sarahlensing: Why do we use `UIApplicationWillEnterForeground ` instead of `UIApplicationDidBecomeActive`?

msmollin: Hmmm I think I misread UIApplicationWillEnterForeground vs DidBecomActive description in the lifecycle docs. Basically I thought that DidBecomeActive would get called when the application became active regardless of foreground or background status. This can happen if your app uses background refresh via silent push notifications. In that scenario, we don't want the GPU to start rendering.

However, re-reading the docs (https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622956-applicationdidbecomeactive), it seems that DidBecomeActive is the appropriate time to restart GPU interaction. Will fix.

MapzenSDK/MZMapViewController.swift
```diff
@@ -456,6 +457,7 @@ open class MZMapViewController: UIViewController, LocationManagerDelegate {
    - parameter seconds: How long the animation should last.
   */
   open func animate(toPosition position: TGGeoPoint, withDuration seconds: Float) {
+    if isInBackground { return }
```
karimnaaji: We would be fine without the `isInBackground` tests since we would let a 'pending' animation until the app is brought back from the background. Now, by adding those, we set this to be a no-op when the app is in the background. I think that it would be important to document that depending on what you end up choosing here (no-op vs pending).

MapzenSDK/MZMapViewController.swift
```diff
@@ -786,7 +795,7 @@ open class MZMapViewController: UIViewController, LocationManagerDelegate {
   }
   
   /**
-   Resets the camera on the current location, as well as the zoom and tilt.
+   Resets the camera on the current location, as well as the zoom and tilt. Can return false if the app is currently backgrounded
```
karimnaaji: Should be in the `returns` block?

MapzenSDK/LocationManager.swift
```diff
@@ -74,7 +74,44 @@ open class LocationManager: NSObject, CLLocationManagerDelegate {
...
+  }
+
+  /**
+   Asks the location subsystem if we're able to receive background location updates. This should be checked every startup because the user can disable it externally in settings.
+
+   - returns: Whether or not the application is able to receive background location updates.
+   */
+  open func canEnableBackgroundLocationUpdates() -> Bool {
+    if UIApplication.shared.backgroundRefreshStatus != .available { return false }
+    if !isAlwaysAuthorized() { return false }
```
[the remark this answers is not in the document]
msmollin: Not sure I see that? `isAlwaysAuthorized()` (which just checks `CLLocationManager.authorizationStatus()`) doesn't necessarily reflect the current `backgroundRefreshStatus` which is a separate setting in iOS apps. Here's Waze, for example:
screen shot 2017-11-20 at 11 09 10 am

MapzenSDK/ApplicationExtensions.swift
```diff
@@ -13,3 +13,9 @@ protocol ApplicationProtocol {
 }
 
 extension UIApplication: ApplicationProtocol {}
```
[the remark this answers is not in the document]
msmollin: Agreed, yep. Will update.

MapzenSDK/MZMapViewController.swift
msmollin: Agreed on documentation for the no-op. Not sure I follow the "pending animation" part? Do you mean on ES 0.9 ? This is still on 0.8 so allowing any calls through to Tangram-es will cause errors?

MapzenSDK/MZMapViewController.swift
msmollin: Ah good call. Updated.

MapzenSDK/MZMapViewController.swift
karimnaaji: Ah yes, this is still pending for 0.9.0 so don't mind it for now. You could consider changing this behavior when updating to the new version.

